PR #2, opened 2021-06-01
Add some guidelines to the issue template

arthuro555:
The com.example one of for displaying the warning about using a unique package name when exporting the game.

The events expanding one is because often users oversee the arrow to expand events and don't understand why they work without seeing the subevents.


4ian:
👍

PR #3, opened 2021-06-06
Add leaderboards example

arthuro555:
A leaderboards example that uses Firebase queries. [I made a tutorial for it as well](http://wiki.compilgames.net/doku.php/gdevelop5/tutorials/leaderboards).


4ian:
Looks great! And thanks for the tutorial :)

PR #8, opened 2021-06-07
Fix top-down-adventure-camera

Bouh:
Game name
delete autosave
format folder name


4ian:
So before we go ahead with this, a few things:

- autosave, good catch. I might change the scripts to avoid this.
- Game name => we use the folder name as a name because it was done like this, but we should be able to read the name directly from the game. What do you think?
- folder name => I'm thinking of removing the convention of writing in kebab-case. We could use any character (including spaces) except for a few ones (namely a `+`, `?`, that would not work in URLs).... or maybe we just keep kebab-case for convention?

Bouh:
> * autosave, good catch. I might change the scripts to avoid this.

👍 

> * Game name => we use the folder name as a name because it was done like this, but we should be able to read the name directly from the game. What do you think?

👍 

> * folder name => I'm thinking of removing the convention of writing in kebab-case. We could use any character (including spaces) except for a few ones (namely a `+`, `?`, that would not work in URLs).... or maybe we just keep kebab-case for convention?

👍 The lambda user never format the name, it can be cool to automatize this name from the game name?
The game.json and the folder have always the same name.

> (namely a `+`, `?`, that would not work in URLs).

What about dot ? My opinion is we should always format the data from the user, a fancy name is always a source or problem...
```My.cO0Iɓ▀.png_trOOI.json``` is a valid file on Windows it's not really pro.

PR #10, opened 2021-06-11
goose-bomberman - Remove a comments that said something wrong

D8H:
It removes a comment about using a group to have an alternative pick list which is false.
It also adds an icon.


Bouh (approved):

4ian:
Sorry for the delay! :)

PR #12, opened 2021-06-17
New example: Random points with min and max distance

LeopoldTal:
Demonstrates how to generate points within a square, circle, and ring (annulus) between two circles.

- [X] Game name ("Random points with min and max distance")
- [X] Author name ("hawkbyte")
- [X] Package name ("com.example.areasampling")
- [X] README
- [x] Thumbnail
- [X] All events unfolded
- [X] Resources cleaned up


Midhil457:
Hi, 

- You should be putting the project inside the example folder, not in the root folder
- If I am correct, the example's name is taken from the json file. So, it might be a good idea to change the json file's name to the project's name
- It would be great if you could shorten the example's name as the long name is causing some texts to overflow in some parts of the GUI, it would also make it search friendly

:)

Midhil457 (changes requested):
Remove the old folder from root folder

LeopoldTal:
> Remove the old folder from root folder

Hm? That's already done. Try `git clean -d`?

Midhil457:
> Hm? That's already done. Try git clean -d?

I guess it was my issues, that folder doesn't seem to exist in the branch

- I saw the thumbnail, and I am not sure if I am correct. but, picture as big as 420x420 might not be required. The platformer example seems to have a thumbnail thats 64x64. Sorry that I didn't mention it in the first review 😅

Midhil457 (approved):
Looks good to go

:)

4ian:
Merged! Thanks both 👍👍

PR #14, opened 2021-06-26
Snap - add a scene for each grid

D8H:
It's the same as this: https://github.com/4ian/GDevelop/issues/2586


Midhil457:
Hi
- You have a useless object "Drawer" in the **PointyToppedHexagonalGridScene** and the **FlatToppedHexagonalGridScene**

- In the **RectangularGridScene**, you can untoggle "Clear the rendered image between each frame" for **drawer** object and also put the grid drawing events under **at the beginning of the scene**

- You have a variable **i** in all the scenes, do remove it if useless or rename to something that describes it's use better

- Also do add a proper package name in the properties section, something like **com.gdevelop.snapexample** might work, it is your choice

:)

Midhil457 (changes requested):
Hi, 

- Remove the drawer object in the PointyToppedHexagonalGridScene and the FlatToppedHexagonalGridScene as these scene don't use it

- Add a proper package name in the game properties panel
- Initialize variables (sorry that I didn't mention it in the first reveiw)

:)

D8H:
>     * Remove the drawer object in the PointyToppedHexagonalGridScene and the FlatToppedHexagonalGridScene as these scene don't use it

Oh, yes, sorry, I forgot this.

>     * Add a proper package name in the game properties panel

ok, I changed it, but I don't see the point as it is only for mobile export.

>     * Initialize variables (sorry that I didn't mention it in the first reveiw)

I think that declaring every variables is not a good idea because then the reader can't tell which variable is used during the lifetime of the scene or not. In this case, it should be a temporary variable when it exists.

Midhil457:
> I think that declaring every variables is not a good idea because then the reader can't tell which variable is used during the lifetime of the scene or not. In this case, it should be a temporary variable when it exists.

That makes sense

Why is there a 1 in the drawing grid event? The 1 makes the grid mess up. I am extremely sorry for not noticing it earlier. Sorry if its getting repetitive

Snap1


:)

Midhil457:
> Also, one of the circle is rotated


I think you forgot to fix this

D8H:
Nice catch, I'm the one that's sorry 😉

Midhil457 (approved):
Looks good now. I think this can be added now. Approved :)

Bouh (approved):

Bouh:
I've made a quick review, good reviews have been done before, nothing to tell, nice work. 👍

PR #18, opened 2021-06-29
Camera forward example by Cat-Robot

Midhil457:
Example by @Cat-Robot
PR for #16 

The example has been reviewed and the needed changes have been made

This example shows how to make the camera lean/face towards the direction of the player, making the path more visible


examples/forward-camera/README.md
```diff
@@ -0,0 +1 @@
+A platformer camera that leans towards the player's direction. Making the obstacles more visible
```
4ian, line 1: "A platformer camera that leans toward the direction the player is looking at. Useful to give more visibility to the player when running through a level."

(careful about your original description, it does not include a period at the end, and you can't start a sentence with "Making the [...]")

4ian:
Thanks for taking the time to review this! :) Added one comment and then good to go :)

Midhil457:
My bad, fixed it.

Thanks

4ian:
Thank you!!

PR #22, opened 2021-07-02
Event clean up #2

Midhil457:
Batch 2

I have some having some doubts if I have should done it n some examples, it would be great if you could give some opinions
- I changed out the assets to a smaller pixel art assets in the keyboard practice example to reduce the size. I was really successful at it as the project size reduced from 389kb to 218kb
- I removed the last two levels in the level selection example as these many levels where not needed for showing how the example works
- And in the count down example, I changed it to use a timer and invert the timer (count down instead of counting up) in the text using `ToString(round(abs(TimerElapsedTime("CountDown")-10)))`, I have some doubts if this will confuse some people

### Move-Camera-to-position
- Use lerp to move camera smoothly
- Change Scene name to "Game"
- Change the layer name from "message" to "GUI"
- Fix spelling mistake in text object

### Update Drag Camera with Pointer example
- Add separate scene that uses the DragCameraWithPointer extension
- Rename Objects
- Change scene name to describe the what it shows
- Increase version number by 0.1.0

### Update Character selection example
- Renamed Player group to PlayerTypes
- Remove events and game mechanics that were not necessary to show the main purpose of the example(Virtual controls, remove 2 enemy type, remove moving platformes, etc...)
- Increment version number by 0.1.0

### Lights
- Increment version number by 0.2.0
- Simplify the example
- Now a room with lights created randomly with some light obstacles
- Remove other objects that are not necessary

### Count Down timer
- Use timer instead of variable and use ToString(round(abs(TimerElapsedTime("CountDown")-10))) to make the text show the timer counting down and not up
- Increment version number by 0.1.0
- Change scene name to CountDown
- Change "GOText" object name to "GameOver"

### Tic Tac Toe
- Remove unused resources
- Rename Project name to English
- Use arrays instead of seperate variables
- Imporve the system to check if the game was won
- Change background colour to white
- Add comments explaining hwo the example works
- Shrink hitbox of blocks as they made clicking multiple blocks at once possible
- Align th blocks better
- Inizialize some variables
- Increament version number by 0.1.0

### Keyboard Practice
- Use an already initialized array instead of creating a structure at the beginning of the scene
- Simplify and update the events
- Use different asset for keys for reducing project size
- Change rendering to support the small assets that are being scaled
- Change font to match the style
- Increment the version number by 0.1.0

### level selection unlock
- Remove the last two levels as these many levels weren't needed to show the working of the example
- Move Reset Memory and Read from memory to separate external events
- Add a clear memory button
- Increment version number by 0.0.1


tristanbob:
You are doing great work!  On behalf of all GDevelop users, thank you for improving these examples.

Bouh:
There is a lot of work here.
It's going to take a little while to check all this out in details.

Globally it is a proper clean-up, my review isn't a complete one, feel free to check behind me.


Since the examples are downloaded the cost is mainly on resources, think to verify to not add more sprite than needed and if you relink a sprite into an object check if the resources haven't changed.

I can see for **level-selection-unlock.json** a new resource file has been added, there are two resources pointing on the same file.
image

There are 6 scenes in this example now, it's correct, for nitpicking because we are at this point of optimization now ^^, the events are the same every scene, an external event can be used.
The event at the beginning of the scene can be reduced into one (sub-event is unnecessary, the action is still require, just move up).

# Keyboard Practice

An idea:  sound of error when the key pressed isn't correct

# Lights 

Nice simplification 👍 
Two resources down, **Turn.png** and **assets\Turn.png.png**

Midhil457:
Thanks for the review, I have done all that you mentioned

> An idea: sound of error when the key pressed isn't correct

I have added it. now it plays a error sound when a wrong key is pressed/clicked. The sound is from kenny assets

The system might not look good, but that the only way I way I could find without any bugs.

Like, you would expect this to work,
NotTheWayToDoIt

But, because there is a "any key" condition, for some reason ignores all the other conditions and plays the sound every time a key is pressed no matter if the key was correct

This is the way I found which works
WayToDoit

:)

4ian:
Seems like this would be ready to go? :)

Bouh:
I've fixed the events for the error sound.
image

Bouh:
Thank you everyone for the reviews and @Midhil457 thx for this initiative!

PR #25, opened 2021-07-04
Remove "Move object with position" example as it is a duplicate of "Move object with physics"

Midhil457:
Remove **Move object with position** example as it is a duplicate of **Move object with physics**

Another Note: The **Move object with physics** example's name is misleading as the example uses forces and not the physics engine. Not to worry, I have already fixed it, (including some other improvements done to the example) in the Event Cleanup 3. I will publish it when I finish everything

:)


4ian:
Thank you 👍

PR #37, opened 2021-07-17
Remove duplicate of "Basic Ai with pathfinding"

Midhil457:
- Removed "Basic artificial intelligence" example as it was a duplicate of "Basic Ai with pathfinding"

- Improve/Update "Basic Ai with pathfinding"
    - One notes, the AI seems to act weird and not consider Wall (a tiled object) as a pathfinding obstacle. Is it because it is a tiled object?


Bouh:
> * One notes, the AI seems to act weird and not consider Wall (a tiled object) as a pathfinding obstacle. Is it because it is a tiled object?

Yes, it seems, I don't know what going on, the collision mask are ok.
Can you revert to classic sprites and compare with the original?

Midhil457:
I tried replacing it with the normal sprite, but it still happens

Bouh:
Even with a rotated Sprite it happens, this should be reported, there is definitively something wrong with Pathfinding and their collisions.

D8H:
I tested the NavMesh pathfinding (https://github.com/4ian/GDevelop/pull/2856) on one of the AI example:
https://games.gdevelop-app.com/game-05f6bf02-0a65-4cc9-bb77-c7c7e0b3c1ce/index.html
It seems to fit well because the tanks can go 360°. Maybe it could be used for the rework of the example when the PR is merged.
It uses hitboxes so tiled objects could work too.

4ian:
Great idea to update this once we have the new NavMesh pathfinding 👍

Midhil457:
I'll make another PR that just removes the duplicate without doing any improvements to the examples. This can be merged or updated when the bug is fixed

:)

PR #40, opened 2021-07-18
Update cannon-ball-with-physics

TheGemDev:
changed The Gem Tutorials to The Gem Dev


Midhil457:
Hi!
I think you forgot to change it here,
gem

I know the ones below are not related to the changes, but I thought it might be a good time to bring it up,
- You can put 0 for immediate rotation to position
  
can1

- Not sure if this is intentional, but you are not checking for mouse press, so it opens a lot of tabs together
can2

:)

PR #60, opened 2021-08-06
Add Magnetic Effect example

ClementPasteau:
Add a game example based on the magnetic extension created here: https://github.com/GDevelopApp/GDevelop-extensions/issues/195


Midhil457:
Hi!,

- There is already a magnet effect example, so this could a) combined into that example as a new scene, or b) replace it

Now here is my review of the example:
- There is no `README.md` file included for the description
- The text in the scene is giving the wrong info of `Click and drag the player to it move around.`, that doesn't work
- This is more related to the extension, the action ` <object> attraction has stopped` sounds more like a condition that an action
- The package name hasn't been filled, also the version is still `0.0.1`, when the example is added it can be `1.0.0` and increment as if there are changes done after
- There can be some comments that explain what the example/extension is
- If I remember correctly, the `.json` file needs to have the same name as the folder. This format is followed by all examples, so it might be necessary
- A small thing, please change the name of the scene from `New Scene` to something that describes the scene better

:)

ClementPasteau:
Thanks for the great feedback @Midhil457 !! Exactly what I was looking for :)

ClementPasteau:
I've merged the 2 scenes into the existing example as I thought they both brought value in their own way!
I also took the opportunity to modify the existing scene to use the new MagneticEffect extension which simplifies the events logic too :)
All the feedback has been addressed, except the "click and drag the player to move it around" -> it works for me as it has a draggable behavior. Was it just a misclick from your end @Midhil457 ?

Thanks for the feedback again! :)

4ian:
Good to go from my side 👍

Midhil457:
Hi!,
There are some old actions use and things that can be done better in the old scene, but I don't think it's the scope of this PR

- For me the magnet effect folder still exists with the .autosave inside, not if it's because it didn't update correctly (Has happened before)

Small thing:
- The project name is `Aimant` which I guess is magnet in some other language, also the author field, and the package name are not filled in
> All the feedback has been addressed, except the "click and drag the player to move it around" -> it works for me as it has a draggable behavior.

It was a mistake on my side 😅  , I thought the player will move towards the mouse when you press down, now it makes sense 

:)

4ian:
> For me the magnet effect folder still exists with the .autosave inside, not if it's because it didn't update correctly (Has happened before)

Autosave files are ignored by git, so as long as it's not in the files in the commit ([visible here](https://github.com/GDevelopApp/GDevelop-examples/pull/60/files)) it's good :) 

> The project name is Aimant which I guess is magnet in some other language, also the author field, and the package name are not filled in

Nice catch! This example was very old and "Aimant" is French for "Magnet" 😅

ClementPasteau:
Good catch, I fixed those and will merge :)

PR #63, opened 2021-08-09
Clean up resources with missing files and add a check for it

4ian:
This removes resources from projects with missing files (that would lead to a download error when opening these examples).
Also add a check for the resource file existence. Otherwise, this creates a bad user experience when opening these examples.

Also fix a wrongly modified README.


ClementPasteau (approved):
Looks nice to me!
Have you managed to check all examples ?
I'm wondering if we could easily add a check in the CI to open all projects and ensure they don't error...

4ian:
This new check is actually checking all examples ;) 

The call to checkProjectResourceFiles is called on all examples, as they are iterated on to build the "database": https://github.com/GDevelopApp/GDevelop-examples/pull/63/files#diff-d69b06da4f18d364844df1fe549525e1f71cfa66a9a80c11b2654fd73e7b7605R256

There was only 3 examples having missing files, so it's fixed for these and for all forever now 👍

PR #72, opened 2021-08-16
Fix typo in Ping-pong readme

Bouh:


Bouh:
I'm also not able to play it with the web app, the code generation crash, it's only on the web app, I've cleared by caches, it persist.

https://game-previews.gdevelop-app.com/1629128780458-117196/index.html

image

Bouh:
Found why!
An event from the desktop isn't available for web app.
@4ian I editing the events and push a fix, can you see on the generation if something can be done? 

Desktop:
image

Web app:
image

4ian:
Good catch. This is an extension that was probably not exposed to the web-app while it should.
We should also probably avoid the error in the code generation, I'll take a look.

Bouh:
Ready to go ✅

4ian:
This will also crash for the web-app ;) If you export the game from the web-app, it won't work because the action is not found.
Two fixes to do on GDevelop (I'm doing it):
- Generate just empty code for unknown instructions.
- Fix the missing extension for the web-app.

Issue #68, opened 2021-08-15
Minesweeper-like game


tristanbob:
~~**Note:** This is a work in progress that is 90% complete.   It needs the logic to clear all neighboring tiles when a user clicks on an empty tile.  I'm asking for help on the GDevelop forum, so please help if you can!~~

## Describe the example

A simple example game similar to Minesweeper.

## Checklist

- [x] My game has a proper name in the game properties. 
- [x] My game package name behind with `com.example.`.
- [x] My game has all events unfolded.
- [x] I've added myself as the author in the game properties.
- [x] I've included a file called "README.md" with a description in proper English, explaining what this example is doing.
- [x] I confirm that this game and all of its resources can be integrated to this GitHub repository, distributed and MIT licensed.
- [x] I've cleaned unused resources.

## Game folder

~~[MinesweeperExample.zip](https://github.com/GDevelopApp/GDevelop-examples/files/6987365/MinesweeperExample.zip)~~

## Playable game

~~https://games.gdevelop-app.com/game-077ea6e2-2977-4e0f-ab4d-36eca9cc038c/index.html~~

tristanbob:
## Version 1.0.0

Thanks to Alaska for contributing a method to reveal empty tiles by placing a "tester" sprite and checking for collisions.  It is quite simple and effective, and a unique way to solve this problem that is normally done via recursion.

This example is ready to be reviewed!

## Project folder

[MinesweeperExample.zip](https://github.com/GDevelopApp/GDevelop-examples/files/6988527/MinesweeperExample.zip)

## Playable demo

https://games.gdevelop-app.com/game-5d61fbc3-abdf-40e3-a7eb-99b63de1e1fe/index.html

Midhil457:
Hi, I checked out the example for a brief second, and I think something is wrong

This is the only layout I get, just in different positions. From the small amount of minesweepers I have played, I think this is not right how it should be
![mw1](https://user-images.githubusercontent.com/73597906/129533062-a954e09b-3e0c-473c-87ac-ae65c15c45cf.PNG)

Bouh:
It's so cool, I'll review this example tomorrow.
I'll also update graphics style with font and colors from Ping pong example recently added for a visual consistency :)

Bouh:
Continue on #74

PR #74, opened 2021-08-17
Add minesweeper like game

Bouh:
Close https://github.com/GDevelopApp/GDevelop-examples/issues/68

image

- [ ] Sometimes Z-order of mines are not enough, mines can be visible.
- [ ] We should wait for the first click before revealing a part of the grid.

If you wanna try it: 
[minesweeper.zip](https://github.com/GDevelopApp/GDevelop-examples/files/6995985/minesweeper.zip)


tristanbob:
Woah, you really kicked this game up two notches with those changes!  Great job and thanks!

tristanbob:
With our powers combined, this example just keeps getting better and better!  :)

## Updates

- Levels now start small and increase in size as the number of mines increase
- No more accidental clicking at the start of a level
- Peeking now shows/hides ALL mines.  Tested with 200 mines on the map.
- UI cleanup of how background object is positioned (and removed shape painter)

## Project folder

[minesweeper.zip](https://github.com/GDevelopApp/GDevelop-examples/files/6997468/minesweeper.zip)

## Playable game

https://games.gdevelop-app.com/game-6f25e622-6b1f-453f-a771-56dc73a2f466/index.html

## Video

https://user-images.githubusercontent.com/8879811/129673061-db448263-2ecb-4c9c-adb3-4f946340c895.mp4

Midhil457:
This is getting better and better! I will try to make a review soon

If you guys are going for consistency. I have a quick mock up design which has a similar design to the breakout and ping pong example. It uses a similar color palette (not the same, as there are not enough colors) 

rect4086

:)

Bouh:
Thanks guys for your help!

@Midhil457 I like the color palette that you suggest! Take care of the whole UI, everything is important.

@tristanbob  - On the video you lose at Level 2 but you click on a number, this wasn't expected, its fixed. (Prefer checking object variable than checking for collisions, it's faster and here we go back on the previous working behavior)
- In Minesweeper function there is a unused function, you wish continue it while Midhil457 try to change the styles or we go without? The game works fine wihtout.

4ian:
Wow! This is getting verry pretty overall! 

Let's not overthink it, but maybe something to be touch screen compatible, like a long press to put a flag, would be great to finish this example?

D8H:
I think that 0 that are only in diagonal should not be reveal. It may be fixed by making the tester hitbox in the shape of a "+". 

There is feature that avoid to kill the flow for quick players: clicking on a number square that has already its neighbor flags set simulate a click on each of its 9 neighbors that are not flagged. In my opinion, once you used this you can no longer play without.

tristanbob:
> @Midhil457 I like the color palette that you suggest! Take care of the whole UI, everything is important.

@Midhil457 UI mockup is good, but I like the UI made by @Bouh better.

> @tristanbob
> 
> * On the video you lose at Level 2 but you click on a number, this wasn't expected, its fixed. (Prefer checking object variable than checking for collisions, it's faster and here we go back on the previous working behavior)

Sorry, I was trying to "simplify" the game by using collision checks, but the object variables work the best.  Thanks for fixing it.

> * In the Minesweeper function there is a unused function, you wish continue it while Midhil457 try to change the styles or we go without? The game works fine wihtout.

Which function is it?  (I don't have access to GDevelop at the moment to test your newest version)

tristanbob:
> Let's not overthink it, but maybe something to be touch screen compatible, like a long press to put a flag, would be great to finish this example?

Enabling the touch interface for mobile devices is a great idea!  Instead of a long press (risky because you might trigger a bomb) how about dragging a flag object from the side onto the tile?

If no one adds this, I'll work on it tonight.

D8H:
> > Let's not overthink it, but maybe something to be touch screen compatible, like a long press to put a flag, would be great to finish this example?
> 
> Enabling the touch interface for mobile devices is a great idea! Instead of a long press (risky because you might trigger a bomb) how about dragging a flag object from the side onto the tile?

Long press will kill the flow. Dragging the flag will be a pain for the player, to much back and forth.
What about a hold to put flag button? It allows instant toggling and the finger is the memory so the player always know which mode is active. It works only with multi-touch though.

4ian:
Happy with a button that you can press with one finger in addition to touching a tile - sounds both safe and efficient :) (no strong feelings though, what you think is the best!).

Happy to merge this template once this is done :) (having touch support is important for most examples)

tristanbob:
> I think that 0 that are only in diagonal should not be revealed. It can be fixed by making the tester hitbox in the shape of a "+".

This is what we are doing already.  Is it not working as you expected?

> There is a feature that avoid to kill the flow for quick players: clicking on a number square that has already its neighbor flags set simulate a click on each of its 8 neighbors that are not flagged. In my opinion, once you used this, you can no longer play without it.

My wife also complained about this missing feature, it must be important!  Can you explain the logic a little more? 

> And I think that the player should not be able to loose on the 1st click. Not having a mine on the 1st revealed square should be a constraint of the level generation.

Is this how the original minesweeper works?  I always thought the first click was supposed to be based on luck.   If we feel this feature is important, we could wait to place mines until after the first tile has been clicked.

tristanbob:
Any thoughts on tile size (currently 48 pixels)?  This site recommends a minimum button size of 48 pixels with a distance of 10 pixels between buttons.  On a game that doesn't use space between buttons, this would translate to a tile size of 58 pixels. 

https://www.voiceandcode.com/our-insights/2019/7/12/ux-best-practices-minimum-target-size-on-mobile-devices

D8H:
> > I think that 0 that are only in diagonal should not be revealed. It can be fixed by making the tester hitbox in the shape of a "+".
> 
> This is what we are doing already. Is it not working as you expected?

The tester hitbox is a square so it tests the 8 neighbors which means that is can propagate on 0 in diagonal.
I propose to make the hitbox the shape of a "+" to only test the 4 direct neighbors to avoid this.

For instance, with this level (* = mine)
```
*10
101
01*
```
At start, nothing is reveled, the player clicks on the bottom left. I think the player should see this:
```
###
1##
01#
```
Because nothing allow to deduce that there is no mine in the center square.

> > There is a feature that avoid to kill the flow for quick players: clicking on a number square that has already its neighbor flags set simulate a click on each of its 8 neighbors that are not flagged. In my opinion, once you used this, you can no longer play without it.
> 
> My wife also complained about this missing feature, it must be important! Can you explain the logic a little more?

If the player put N flags on the 8 neighbors of a square marked by a N, the neighbors without a flag must be safe (or the player made a mistake). So to avoid to click on each of the safe neighbors the game allow to say "hey, I'm confident of my flags. Reveal the neighbors of this square." The result is the same as if the player clicked each neighbor one by one.
If there is less neighbor flags that the number on the square, it does nothing to avoid bad input.
In the Windows version, it was click left + click right, if I remember correctly.

> > And I think that the player should not be able to loose on the 1st click. Not having a mine on the 1st revealed square should be a constraint of the level generation.
> Is this how the original minesweeper works? I always thought the first click was supposed to be based on luck. If we feel this feature is important, we could wait to place mines until after the first tile has been clicked.

I'm not sure but I can't remember loosing at first click even on hard.

tristanbob:
Tonight I'll see if Bouh's latest version still has the correct "+" shaped hitbox that I created.

Todo:
- [ ] Verify tester hitbox is still "+" shaped
- [ ] Make first click always safe (place mines after the first click).  See screenshot below.
- [ ] Allow players to click on revealed numbers to reveal unflagged tiles (but only if the number of flag neighbors equals the number of mines).  UPDATE: I just tested this on Minesweeper and they don't do this "safety" check.
- [ ] Add touchscreen button "Hold to place flag" to enable multitouch placement of flags
- [ ] Consider increasing tile size 

I downloaded the Windows 10 version of Minesweeper and quickly learned that the first click is always safe.  :)
Screenshot 2021-08-17 153032

tristanbob:
> [minesweeper.zip](https://github.com/GDevelopApp/GDevelop-examples/files/6998110/minesweeper.zip)

@Bouh - It seems you did not base this version on my last one, so it is missing a lot of the improvements.  I'm going to keep going from my last submission.  Is there something in this version that I should include in my version?

Also, @D8H was right, the collision mask on my "+" tester sprite was set to the default square.  Is this because I was trying to use a concave shape, so it reverted to the default mask?  I noticed the "concave" warning tonight so I changed the shape to a concave diamond, it should have the same result we want.

tristanbob:
I didn't get done as much as I wanted, but it is still progress.

## Version 0.8.17 (today's date)

- See updated items on  "To do" list above.
  - Verify tester hitbox only touches 4 neighbors (now using a diamond shape)
  - Make first click always safe (place mines after the first click). In fact, no mines should be on neighboring tiles also.

## Playable demo

https://games.gdevelop-app.com/game-5d5c8b6b-761a-4624-85ea-5a0058e50074/index.html

## Project folder

[minesweeper-0.8.17.zip](https://github.com/GDevelopApp/GDevelop-examples/files/7004632/minesweeper-0.8.17.zip)

Midhil457:
> @Midhil457 UI mockup is good, but I like the UI made by @Bouh better.

Yes, I am trying to bring both the UI and the minefield into a similar style. The UI is pretty similar

I am also experimenting with making it able to resize, and add anchoring to the UI

Here is how it looks currently (The blur for the backdrop is intentional, but I am still experimenting with it)
mines

Bouh:
I've found another issue in your last version, place a flag, use left click on it, MineCount texts are shown after.
I've fixed it in the version 0.8.18 bellow.

https://user-images.githubusercontent.com/1670670/129868044-a965a527-39da-47eb-b121-5be93e9b6d2a.mp4

[minesweeper-0.8.18.zip](https://github.com/GDevelopApp/GDevelop-examples/files/7005862/minesweeper-0.8.18.zip)

(Adding a build version is a good thing, it avoids unintentional rollbacks 👍 )

Midhil457:
I have completed the UI Update, with some other changes too

- I don't have the permission to commit to this branch, so I have made another branch `minesweeper`. If possible please merge both of them or I can provide a zip
##Changes
- Update UI
- Made the game responsive (The buttons are still having some issues as they are used in multiple places and have to be adjusted according to it)
- Add Audio (Place flag, remove flag, bomb)
- Clean some events (UI related)
- Center camera on the minefield
- Add Tags to objects
##Notes
-I have added the audio for placing flag, removing flag, bomb. I wanted to also add a digging sound effect when clicking on the field. but, the folder was acting weird and I wasn't able to find a good one. If someone can, feel free to add a sound for that


mine1
mine2

It's now responsive
mine3

Bouh:
> * I don't have the permission to commit to this branch, so I have made another branch `minesweeper`. If possible please merge both of them or I can provide a zip

Let continue with a **zip** file for now.

I'm ok with the new ending message, but the buttons are non-conventional, I have mixed feelings about those.
image

If you add text with the buttons it's better. Like 
image

Note this previous button doesn't work in your previous playable link.

You have a mixed feeling about a blurry backdrop, I prefer without blur. However, this has not real importance.

Bouh:
Also for the font, it looks a bit blurry, if you want you can double the font size, and apply a rescale of =0.5 with events on it.

4ian:
I know it would look better, but I have mixed feelings about applying "hacks" like this in examples, that are supposed to be straightforward and as simple as they can be.

Midhil457:
Updated it with all the suggestions

- Removed the blur from the backdrop
- Added text in the UI_Postcard (There are some issues with alignment)
- Update the thumbnail
Playable game: https://games.gdevelop-app.com/game-3461eca8-98d8-4a5f-93dd-76e39067cdba/index.html
mine6

Thumbnail:
thumbnail

:)

Midhil457:
Updated it with all the suggestions

- Removed the blur from the backdrop
- Added text in the UI_Postcard (There are some issues with alignment)
- Update the thumbnail
Playable game: https://games.gdevelop-app.com/game-3461eca8-98d8-4a5f-93dd-76e39067cdba/index.html
mine6

Thumbnail:
thumbnail

:)

D8H:
The 2 buttons on the top only work at the game over.

I'm a bit surprise that the game end when all the mine are flagged. I thought the player needed to reveal everything.

4ian:
> Is there really a point of the 1st levels? I know it's just an example but the game is not fun as a puzzle and is much better as a time attack game. I think that only one level but with a best time would take less events and be more fun.

I do think it's interesting :) I'm not a pro player at this and I appreciate "learning" the rules (not everybody has played much to it, especially younger people I would say ;)).

Also everyone, let's try to call this "done" at some point, or at least push a first version to the examples ("done is better than perfect").
Remember that I really want this to have touch support => we can't ignore mobile, it's 50%+ of the game exports in GDevelop, almost all examples should work on a touchscreen.

Midhil457:
> The 2 buttons on the top only work at the game over.

Fixed, thanks!

- Fixed the buttons on on the top not working
- Removed the old assets
- Remove the SceneChange variable as I wasn't able to reproduce the issue it was trying to fix. Maybe because I changed out "Mouse Pressed" with "Mouse Released" in some places

Playable game: https://games.gdevelop-app.com/game-0f63f092-0008-40cd-a635-0231d55cef70/index.html
Zip:
[minesweeper.zip](https://github.com/GDevelopApp/GDevelop-examples/files/7014774/minesweeper.zip)

A question not related to the example, How long do exported online games last on the server? I had made many exports, found an issue the next second, made another. So, there are many outdated exports of this example that I have exported, those could be removed from the server.

:)

D8H:
That not really important but the player can cheat with flags. For instance, in the 1st level you can flag each square until you found the mine. I think it can be used at the end of a level too if the player don't know where the last mine is.

I think it's because of this:
> I'm a bit surprise that the game end when all the mine are flagged. I thought the player needed to reveal everything.

tristanbob:
> That not really important but the player can cheat with flags. For instance, in the 1st level you can flag each square until you find the mine. I think it can be used at the end of a level too if the player don't know where the last mine is.
> 
> I think it's because of this:
> 
> > I'm a bit surprise that the game end when all the mine are flagged. I thought the player needed to reveal everything.

I agree we should change that behavior.  The current win condition is just making sure that all mines have flags on them.  This should be changed to something like:  

```
ForEach(Tile)
  Condition:  Tile.Variable(Revealed) OR Tile.Variable(HasMine)
  Action: Add 1 to WinCounter

Condition: WinCounter = Count(Tile)
Action: Set Variable.GameWon to TRUE
```

D8H:
> I got the time to add the UI for the mobile controls. So, currently it is anchored, on a separate layer.

For the hold to flag button, I was thinking of only one button:
* pressed = put flags with an other finger
* released = dig with an other finger

About the ergonomic design, try holding your phone in landscape with one hand, press and release a imaginary button while the other hand is free to point on the board. What is the more comfortable position for you?

I think that the button should be at the top left for right-handed persons and top-right for left-handed person. It will be very hard to make it beautiful that way though.

Midhil457:
The buttons are meant to be ratio buttons. I looked up the microsoft minesweeper for windows phone, and that's how it done for it. It seems simple and useable at the same time, so I went with it

D8H:
> 
> 
> The buttons are meant to be ratio buttons. I looked up the microsoft minesweeper for windows phone, and that's how it done for it. It seemed simple and useable at the same time, so I went with it

Ok, about the title what do you think about:
* Super Green Mine (as a ref to the 5th Element, it could be read as "My precious game of mine" and to remind the graphics)
* Super Green Mined (to try to try to sound like "mind" as it's a refection game)

Midhil457:
> Ok, about the title what do you think about:
Super Green Mine (as a ref to the 5th Element, it could be read as "My precious game of mine" and to remind the graphics)
Super Green Mined (to try to try to sound like "mind" as it's a refection game)

I will go with Super Green Mine, it is similar enough to know what it is. But different enough to not get sued

tristanbob:
> Also everyone, let's try to call this "done" at some point, or at least push a first version to the examples ("done is better than perfect").

Can we have a few more iterations, this is fun!  :)

Here is my short-list that I'd like to see done:

- [ ]  Make first click always safe.  See my latest zip file, I used a larger tester sprite that touches all 8 neighbors. If a tile collides with that, I add Tile.Variable(StarterTile).  Then I place mines and avoid those tiles.
- [ ]  Allow players to click on revealed numbers to reveal unflagged tiles
- [ ]  Add support for mobile devices (looks like @Midhil457 just added this)
- [ ]  Change win condition to require that all tiles are revealed or have bombs (see my comment above)

Midhil457:
Just a suggestion, if we are going to have a few more iterations, starting a github repo might help fast things up and merge all the contributions. This repo can be kept for future improvements too

> Add support for mobile devices (looks like @Midhil457 just added this)
I wasn't able to add the logic, so currently only the UI works. I tried adding it, but I wasn't able to figure it out without some refactoring

D8H:
> Just a suggestion, if we are going to have a few more iterations, starting a github repo might help fast things up and merge all the contributions. This repo can be kept for future improvements too

This is a good idea. I opened a discussion about this: https://github.com/4ian/GDevelop/discussions/2947

4ian:
If mobile support is really too complex, let's go ahead with this version?

Midhil457:
If we are going with the current version, I can remove the partially completed mobile support. And someone could look at it another day and see if they can implement it

Bouh:
We can go with the current version.
We will be able to add mobile support in the next few days.

Can you send the latest version in a zip, I'll take care of fast checking and merging.

Midhil457:
The latest version without the mobile UI is https://github.com/GDevelopApp/GDevelop-examples/pull/74#issuecomment-901929804. The one only added the mobile support

Bouh:
Let's go with the "desktop" version.
We can open a new Issue or PR as you want for adding a mobile support.

Here how is the game on my phone.
Capture d'écrans_20210825-151257

We have here a bug from GDevelop on Panel Sprite I take care of report this bug.
Thanks, everyone for your contribution and your knowledge!

Bouh:
Test for checking if resources are missing is too strict, the test is stuck because `flag.png` is not found, but it exists as `Flag.png` in the folder.

4ian:
The test is valid, it will fail on Linux and possibly in the web-app. The
file should be renamed :)

PR #80, opened 2021-08-28
Fix Anchor on left LoseZone for small screen

Bouh:
Fix the anchor position of left LoseZone (white object) for small screen, see before:
image


4ian:
Nice!

PR #88, opened 2021-09-04
Remove Basic Artificial intelligence example

Midhil457:
It was a duplicate of `Basic ai with pathfinding`

#37


4ian:
Thank you!

PR #105, opened 2021-09-18
remove splashscreen example

Midhil457:
- The newer updates brings customisation to the splashscreen. So, there is no need for an example for that in my opinion

EDIT: I accideently commited a commit to this branch, I have reverted :sweat_smile:


4ian:
Agree this is not needed anymore, we can have another one later if we want to demonstrate a custom one 👍

PR #118, opened 2021-09-23
Updated dungeon generator example

tristanbob:
Sound files changed and all have proper licenses.
Added dinos and treasure hunt to cave dungeon.
Added juice and polish.


Midhil457:
I did a quick look, and I think the room amount, room size should somehow to limited
Or, things like this happen
dun1

- Also, I am not sure, but I think there are some unused extension, do remove them (if any)

I'll try to do a complete remove soon (maybe this friday or weekend)

:)

4ian:
Thanks @Midhil457 for the review! 👍 I suggest I merge now to ensure the new properly licensed sound files are deployed, and your feedback can be handled in a second update in another PR :)

tristanbob:
> I did a quick look, and I think the room amount, room size should somehow to limited
> Or, things like this happen
> dun1

Great observation!  I was aware of this issue, but you gave me an idea of how to fix it.  I prefer to fix the extension rather than simply tweak the example sliders so we can make the extension more resilient to user input.

The current room placement strategy is this:

1) Randomly place the desired amount of rooms (using the specified sizes) inside the map area.
2) If rooms are touching try moving them 1 room distance in a random direction
3) Repeat the second step until the max attempts value is reached.

The way to fix this is by adding a 4th step:

4) After reaching the max attempts value, if rooms are still touching then delete them one at time until there are no more collisions.

Does that sound like a good solution?  Do you have any other ideas?  I think a binary space partitioning strategy might be better, so that could be a future option.

> * Also, I am not sure, but I think there are some unused extension, do remove them (if any)

It is possible, but I did use a lot of extensions (I like to show users how they can be leveraged to make advanced mechanics with very little effort.

> I'll try to do a complete remove soon (maybe this friday or weekend)

If you do review this, can you see why I was not able to add a white outline effect on the dinos?  (It might related to one of the effect bugs that will be fixed in the next release)

Thanks @Midhil457!

PR #122, opened 2021-10-02
Rename rectangle movement folder name

Midhil457:
The folder name was not same as the json file's name

The folder's name was `rectangle-movement-example`
now it's `rectangle-movement`
Other examples follow this pattern (No `example` in name)


Midhil457:
The json used a `_` to seperate words, changed it to use `-`

PR #125, opened 2021-10-03
[BimBam] Fix bubble collision at low frame rate and make it touch devices ready.

D8H:
* The shot bubbles are separated from the board bubble before snapping to avoid to snap inside an existing one.
* The aim angle is changed instantly at mouse click to handle touch devices.


4ian:
Thank you!

PR #129, opened 2021-10-05
Add 3 thumbnails.

D8H:
And make one 128 pixels.


4ian:
Thanks!

Bouh:
Thank you Davy!
As far I know there is no requirement on the size.
So are we go 🚀

PR #153, opened 2021-11-07
Add New example label to issue template

Midhil457:
Changed package name standard to `com.gdexample.<project name>`.Done so that the examples don't accidentally conflict with the games made by users (Which have `com.example` as default) and also to be able to differentiate the examples.

Also added label `New example`

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-example.md
+++ b/.github/ISSUE_TEMPLATE/new-example.md
@@ -2,7 +2,7 @@
 name: New example
 about: Submit a new example to be added to the repository
 title: New example
-labels: 'New example'
+labels: '✨ New example'
 assignees: ''
 
 ---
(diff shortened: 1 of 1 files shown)


4ian:
Hey! As pointed in a PR, I'm not sure we actually want to do this? 
Examples will be started and reused by users - so it means that if they have com.gdexample, that's what the user will end up doing. We actually already have warnings in the app about com.example - and it will be rejected by app stores so we're 100% sure nobody will keep using them.
If we use com.gdexample, we'll actually end up having people using this in the app stores for sure - and there won't be any warning :/ 

So I would recommend actually going back to com.example for everything?

Midhil457:
That’s true and better. My bad 😅 I thought the change was fine as no comment was said against it in #132 (In which I changed them for all the examples) I should have looked for proper comments before going ahead because nothing was said against it. I will surely revert the changes in the coming weeks :)

4ian:
No problem, we can surely make a global find/replace to change all "com.gdexample" to "com.example" in the files of the repository : )

Midhil457:
I reverted the package name change in this PR. The label would be useful, so can it be merged will the label added?

4ian:
Should be useful, but the exact label name including the emoji must be used ;)

PR #156, opened 2021-11-08
Add long description to Virtual Joystick example

Midhil457:
Add a short note on how virtual joysticks can be useful
Adds credit to kenney assets for the art (thought crediting was important). Also mentions that it can be found on the GDevelop asset store


4ian:
Made some tiny changes, thanks!

Issue #149, opened 2021-11-02
New example: Fixed Camera Transitions [Using Tweens]


Elairyx7301:
## Describe the example

This example showcases fixed screen transition. Whenever the player moves outside of the current frame, the camera smoothly moves one room in that direction.
A rather easy but sweet event prevents the player from leaving the scene when a transition is happening. Even though this might be most suitable for a Top-Down game, this example shows that it can also find application with other game styles as well.
This is just a template and can be modified freely to fit the personal needs or match the game's style and/or atmosphere.

## Checklist

- [x] My game has a proper name in the game properties. 
- [x] My game package name behind with `com.example.`.
- [x] My game has all events unfolded.
- [x] I've added myself as the author in the game properties.
- [x] I've included a file called "README.md" with a description in proper English, explaining what this example is doing.
- [x] I confirm that this game and all of its resources can be integrated to this GitHub repository, distributed and MIT licensed.
- [x] I've cleaned unused resources.

## Game folder
[FixedMoveTransitions.zip](https://github.com/GDevelopApp/GDevelop-examples/files/7461229/FixedMoveTransitions.zip)
![image](https://user-images.githubusercontent.com/83909121/139862428-0ce20b40-6c53-4730-8e7d-1389caf9a8a9.png)

Midhil457:
Hi, here is my short review,
- There are broken resources in the resource manager, please fix them
- Hide is not a good name for a object group. Maybe `HiddenObjects` is a better naming
- `placeholder` might be better than `helper` for the resource name
- `Xaxis` and `Yaxis` is a better name than `horizontal` and `vertical` in my opinion as it is better to follow directional names used by the engine (to keep it simple)
- The package name needs to be `com.gdexample.<project name>`.  The format was recently changed so that the examples don't accidentally conflict with the games made by users (Which have `com.example` as default`) and also to be able to differentiate the examples
- The `README` could get some changes. The short description should explain *what the example* does and not *how the example does it*. The `Topdown adventure camera` does the similar thing with the camera, you could look into that. Also for adding a long description, leave a line between the short description and it should be considered as the long description
- I don't think `Fixed Move Transition` is the right name for this example. Maybe `Room Camera`? I don't have many suggestions
- The project is very lagging for me. Might be because of the duct tape setup I am using so that I can do the review, but the other examples do seem to be a bit more smoother

:)

Elairyx7301:
@Midhil457 Thank you for that big review!

[...]

PR #168, opened 2021-11-19
Add Platformer Room Camera

Midhil457:
### Description
This example shows hows to move camera in between rooms using tweens.

In this example, the camera will smoothly move into the next room when the player leaves it in any of the cardinal directions. The examples also has the additions that the player will get an additional jump when jumping up into a transition and will be kept within the current frame when the transition is in progress.

### Details
Author: Elairyx7301
Submission: #149

139862428-0ce20b40-6c53-4730-8e7d-1389caf9a8a9


4ian:
I see you did a review already, so happy for you to merge this!

4ian:
I've just tried the example and it's lagging a bit (~20-30 fps) on my computer, which is weird. Is this also the case for you?

Midhil457:
> I've just tried the example and it's lagging a bit (~20-30 fps) on my computer, which is weird. Is this also the case for you?

I did have some lagging issue on another older computer, but thought it might be bacause it was a 32-bit device. But, I didn't experience any major issue s on my regular computer

I have cut down on coins and removed the effects from them to try fixing it, And also set the max fps to 0 (Not that I think removing the max fps will fix the issue) could you test it out now?

4ian:
Running well - not sure if it was an overload of something or just my computer.
Looks great then, good to go 👍 I let you squash and merge? Don't forget to set a title like "Add Platformer Room Camera example (thanks Elairyx7301)", as the commit title (and description) is going to the changelog :) 

image

image

Midhil457:
Thanks for letting me merge! Merged successfully!

PR #178, opened 2021-12-02
Revert package name

Midhil457:
Revert package name prefix from `com.gdexample` to `com.example.`

A review would be appreciated as I used a python script to find and replace the package names and wasn’t a manual work where I could make sure it was done right. Though as far as I checked, everything looks fine

Sorry for the mistake caused :)


4ian (approved):
Looks all good to me! Thanks for changing this :)

PR #187, opened 2021-12-08
Add description to starters

AlexandreSi:
To display description for the starters that will soon be in the spotlight on the new homepage


examples/space-shooter/README.md
```diff
@@ -1 +1,3 @@
 A beautiful, retro side-scrolling shooter where you must defeat incoming enemies with your mecha transforming spaceship. Huge boss included!
+
+Discover how to organize your project so that, even for an ambitious project, you can keep a clean and tidy creation environment! Check out how this template uses behaviors made with love by the community and also its very own function for dedicated logic.
```
ClementPasteau, line 3: project twice, maybe replace the beginning by "Discover how to organize your ambitious project, by keeping a clean tidy creating environment!"

examples/particle-effects-demo/README.md
```diff
@@ -1 +1,3 @@
 A demo of various high quality particle effects (fire, magic, snow, rune spell...) that you can try and use in your game.
+
+The devil is in the detail and this little detail can often be found in particles! Explore the possibilities GDevelop offers to make a visually stunning game with the Particle Emitter object.
```
charburgx, line 3: "Devil is in the detail" the last "detail" should be plural!

I'd change "can often be found in" to "is often found in"

charburgx (commented):
Here are some thoughts!

examples/isometric-game/README.md
```diff
@@ -1 +1,3 @@
-Example of making an isometric 2D game.
\ No newline at end of file
+An example of a 2D isometric game. If you ever dreamed of creating you own version of Civilization or Warcraft, this example is for you!
+
+Discover what you can accomplish in an isometric world with great and smooth animations. Add as many blocks as you want and stretch the limits of exploration to your imagination!
```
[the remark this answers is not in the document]
D8H: This example is only a demo for character movements in an isometric viewpoint. Giving big names of strategy games is very misleading. By the way, these kind of games would be very difficult to do with GDevelop.

Personally, in general, I don't like enthusiastic formulations. It feels like an attempt to manipulate me and it's it usually poorly informative. For me, it kills trust and is actually very cold and impersonal.

The reader already downloaded the engine, maybe an informative and concise description would be more effective to let them find the example they need.

examples/downhill-bike-physics-demo/README.md
```diff
@@ -1 +1,3 @@
-A fun demo, using the Physics Engine to build an advanced simulation of a bike, with a person pedaling on it. Try to cross the finish line without falling and as fast as possible!
\ No newline at end of file
+Try to cross the finish line without falling and as fast as possible! And remember to pedal safe!
+
+Start from this template and use the Physics Engine to build an advanced simulation of a bike, with a person pedaling on it! You will discover how to join objects, apply basic physics behaviors to your scene and change your camera settings to adapt to what's happening during the game.
```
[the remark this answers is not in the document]
AlexandreSi: We tend to use lots of commas in french, I have to lose this habit in english!
                                                                    ⬆️ 😅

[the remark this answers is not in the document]
AlexandreSi: The question of the tone is complicated. To be consistent with the rest of the app (in which there is no contraction), I'll leave `You will`.

examples/isometric-game/README.md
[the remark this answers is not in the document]
AlexandreSi: I ended up using Wesley's proposal that felt less superlative.

Midhil457:
As the icons are now 16:9 won't it make the icon look wird in the example list/examples library? Or was the examples list/library removed? Or will the starters not be visible in the example library and only be accessable from the carousel?

AlexandreSi:
> As the icons are now 16:9 the starters won't it make the icon look wird in the example list/examples library? Or was the examples list/library removed? Or will the starters not be visible in the example library and only be accessable from the carousel? Or will the start icons be shown differently than other example icons in the list?

Great question, I was just thinking about it! I made sure that 16/9 images will display correctly in the example list (see https://github.com/4ian/GDevelop/pull/3340/commits/534817a8e1d1846f960608dbd3784d6bf9f7cbc5)

So, to sum up, users without the new version will see square icons with 16/9 images fitting in: not great but I think it's better having the new homepage with great thumbnails, than avoiding that.

Of course, we could have an API that returns thumbnails of different ratios, but for the moment it's a bit overkill

Midhil457:
Maybe have another `preview.png` image that will be used for the carousel. It can only be required by the starters. And the `thumbnail` will be shown else where. This would also give an option to later have a preview image of the example in the description dialog

AlexandreSi:
> Maybe have another `preview.png` image that will be used for the carousel. It can only be required by the starters. And the `thumbnail` will be shown else where. This would also give an option to later have a preview image of the example in the description dialog. And also give way to have the examples in other places too as there are two image types to choose from

Well well well, it's a great idea! I'll update my PR https://github.com/4ian/GDevelop/pull/3340 and this one

4ian (approved):
Approving now, feel free to merge when ready. The new logic for the preview.png file looks good!

examples/isometric-game/README.md
```diff
@@ -1 +1,3 @@
-Example of making an isometric 2D game.
\ No newline at end of file
+An example of a 2D isometric game. If you ever dreamed of creating you own version of Civilization or Warcraft, this example is for you!
+
+Discover what you can accomplish within an isometric world using smooth animations. Add as many blocks as you want to create whatever you can imagine!
```
D8H: Even without superlatives, making it sounds like it will allow someone to make something like a AAA strategy games when the example is a very early **adventure game** prototype is misleading if not dishonest.

The best feature of this example is the art done by Mickael. The tile positioning doesn't allow to "Add as many blocks as you want" without having headache because the tiles doesn't align on the grid. I have a version that fix that I could share.

What about a description strait to the point like this:
```
This example shows how to build an isometric word and handles character movements.
```

4ian: Agree that the reference to AAA strategy game is not the best thing here, because it's far from this and the example is very simple, while making isometric games is still fairly complex.

Thanks for challenging this!

AlexandreSi: I admit I've been a bit cavalier with that!
I'm always amazed by what the extension the community manages to create, I wouldn't be surprised if someone create an extension that allows you to create civilization-like games easily.

PR #190, opened 2021-12-10
[Isometric] Allow to use the grid to add new tiles.

D8H:
Changes
* The assets are shifted to make them align on the grid. This allow an easy positioning of new tiles.
* The map is bigger than the original. This is what I used for NavMesh path finding examples.

The sets of icons for Android and iOS take more than 10% of the project weight. Those icons are useless imho, what do you think about removing them to save 10% of the downloading time?

IsometricGrid


4ian:
> The sets of icons for Android and iOS take more than 10% of the project weight. Those icons are useless imho, what do you think about removing them to save 10% of the downloading time?

Yeah that's a bit too much, at the same time it gives a nice touch to the game when exported so I'm not sure. Hopefully with the increase in parallel requests we should not feel too much this size when downloading the project?

I'm merging this now because the improvements are very nice - and we can think about icons again for another PR.

PR #196, opened 2021-12-12
Add an island generator example

D8H:
## Describe the example

This example shows how to generate top-down islands using the "Marching Squares" and "Noise Generator" extensions.

Try it on Itch.io: https://d8h.itch.io/island-generator

IslandGenerator-Screen

## Checklist

- [x] I've followed all of [the best practices](http://wiki.compilgames.net/doku.php/gdevelop5/community/guide-for-submitting-an-example).
- [x] My game has a proper name in the game properties. 
- [x] My game package name behind with `com.example.`.
- [x] My game has all events unfolded.
- [x] I've added myself as the author in the game properties.
- [x] I've included a file called "README.md" with a description in proper English, explaining what this example is doing.
- [x] I confirm that this game and all of its resources can be integrated to this GitHub repository, distributed and MIT licensed.
- [x] I've cleaned unused resources.

## Game folder

* Project: https://www.dropbox.com/s/eybx7ytzctk8bq1/marching-squares-island-generator.zip?dl=1


Bouh (approved):

D8H:
> If I understand very simply 😅 Some circles are drawn with MarchingSquares extension. The normals of these circles are set up with the data from a noise function.

It uses the values of 2 fields to constraint the generation with an upper and lower bounds.

> It's not easy to understand, maybe above the events, you could add a comment that describes how it works?

I will add a section in the doc for island generation. There is already long comments in the example. I'm not sure adding more will help but illustrations from the doc will, I hope.

> And the roles of each object.

Some are already explained where they are used, but I can add a comment at the start to give an idea of what they are used for.

> One annoying thing is to find out what the actions and conditions related to the extensions do. The descriptions of each function are basic.

For instance, for this action, the description is "Merge with another field". Do you suggest to replace it by something like: "Apply a given operation on every value of this field using the value from the other field at the same position."?

MergeAction

I'm not sure if this will be easier to understand. I prefer that users refer to the doc because it's hard to visualize without any illustrations.

> I read your documentation about [Marching Squares,](https://wiki.gdevelop.io/gdevelop5/all-features/extensions/marching-squares) it help a lot, we could add this link in the comment on top of the events.

There is already an help button for that. It doesn't need to be copy-paste and it stays up to date if the doc is moved.
It might not be a reflex for users but, in this case, this is a GUI issue.

Capture d’écran 2021-12-07 202600

Bouh:
> For instance, for this action, the description is "Merge with another field". Do you suggest to replace it by something like: "Apply a given operation on every value of this field using the value from the other field at the same position."?
> 
> MergeAction
 
Yeah ! As a long description. The sentences from the events is fine.
The idea I was in mind was to have a better technical description because "merging" term is too simple I think it doesn't explain much. 

> I'm not sure if this will be easier to understand. I prefer that users refer to the doc because it's hard to visualize without any illustrations.

Sure!

D8H:
Ok, I'll add details on the descriptions and see if I can add more comments.

4ian:
Awesome doc! I've moved the page to https://wiki.gdevelop.io/gdevelop5/extensions/marching-squares (the convention is to use gdevelop5/extensions/extension_name) (and the reference is generated in gdevelop5/extensions/extension_name/reference).

Issue #200, opened 2021-12-26
New In-App Purchase example


Jurfix:
## Describe the example
This is a reworked In-App Purchase example, since the official example does not work (withdraws money from the card, but does not add a reward), but this one works, I tried to fix the official example, but I still did not understand what exactly the problem is most likely due to aliases or loops or "trigger once".  In any case, this example works and I even integrated it into my application in the market https://play.google.com/store/apps/details?id=com.jurfixgame.mkpp , the inoperability of the official example was also confirmed @TheGemDev  and filmed a video with a working example https://www.youtube.com/watch?v=JV7H6l_ut5U&t 

## Checklist

- [x] I've followed all of [the best practices](http://wiki.compilgames.net/doku.php/gdevelop5/community/guide-for-submitting-an-example).
- [x] My game has a proper name in the game properties. 
- [x] My game package name behind with `com.example.`.
- [x] My game has all events unfolded.
- [x] I've added myself as the author in the game properties.
- [ ] I've included a file called "README.md" with a description in proper English, explaining what this example is doing.
- [x] I confirm that this game and all of its resources can be integrated to this GitHub repository, distributed and MIT licensed.
- [x] I've cleaned unused resources.

## Game folder
[In-AppPurchaseExample.zip](https://github.com/GDevelopApp/GDevelop-examples/files/7777181/In-AppPurchaseExample.zip)

Bouh:
Oh, great to see that one works better. I'll look into it and do a review this week.

Bouh:
## Review

Steps comments are very nice 👍 

- [ ] More all resources in asset folder, (becareful before moving the asset you have two assets with the same name `NoAds-1.png` but with a different visual)
- [ ] Clean unused resources, assets/Tile-12.png is unused
- [ ] No need for all the icons, if you don't know how to clean it I can do it later.

 @arthuro555 any concern on this example?
 Trigger once and loops has been removed, and as said Jurfix it works better without.

arthuro555:
Looks fine to me 👍 

Little derail though, the "Mark purchase as delivered" should be called last, as otherwise you are marking the delivery as complete before actually completing the delivery, breaking your contract with Google Play/Apple App Store. It is pretty minor and arguably unnoticeable but let's still give the good example by putting it after the delivery of the purchase (last action in action list).

4ian:
Thanks @Jurfix for this fixed example and @Bouh and @arthuro555 for reviewing! 
[...]

PR #205, opened 2021-12-30
Update IAP example to 1.1.0

Bouh:
Mainly a simplification of this example.
And I added a basic thumbnail.

From https://github.com/GDevelopApp/GDevelop-examples/issues/200


4ian (approved):
Thanks!

PR #246, opened 2022-04-14
[PingPong] Use new touch functions from GDevelop 5.0.132

D8H:
This should not be merged before the release of GDevelop 5.0.132 as it won't work with the current version.

There is no control over the ball direction. The classic pong uses the speed of the paddle. It might be too tiny on a smartphone to do the same. Otherwise, what was done for the breakout example could be applied here.

* Build: https://liluo.io/d8h/ping-pong
* Project: https://www.dropbox.com/s/anxnlu8tirg1w3u/multitouch.zip?dl=1

It can be reviewed with the nightly:
* https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Nightly-Builds-and-continuous-deployment.md

### DIFF
--- a/examples/ping-pong/ping-pong.json
+++ b/examples/ping-pong/ping-pong.json
@@ -16,8 +16,9 @@
     "scaleMode": "linear",
     "sizeOnStartupMode": "adaptWidth",
     "useExternalSourceFiles": false,
-    "version": "1.0.0",
+    "version": "1.0.1",
     "name": "Ping Pong",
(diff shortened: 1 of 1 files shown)


Bouh:
LGTM 🚀 

For the index variable, we need it because we are in a loop, if we could add an expression to get the index of the loop with a built-in expression from the engine it could be easier.
A loop logic isn't friendly to create, and speaking index to a newbie don't help.

Bouh (approved):

PR #256, opened 2022-04-21
Add Runner-like game example

fannieyan:
image

You can try the game here : 
https://liluo.io/lyve/run-dino-run
(Note that the input is still displayed when the leaderboard loader is displayed ; it shall be solved through the engine instead)

## New iteratio
[run-dino-run.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8555449/run-dino-run.zip)
- Improved touchscreen game experience

## Old
[run-dino-run.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8554871/run-dino-run.zip)
- Fix obstacle spawning in the middle of the screen
- Add touchscreen-specific instructions

[run-dino-run.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8553482/run-dino-run.zip)
- Add sand and bonus particles
- Fix spacing and out of screen sprite creation
- Add custom points for easier object placement

[run-dino-run.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8541386/run-dino-run.zip)
- Added a title screen
- Make player name persist through session
- Solve naming issues (scene, event groups)
- Remove some redundancies
- Reworked some sprites
- Fixed inverted buttons
- Simplified expressions for ObstacleSpeed and ObstacleSpawnDelay incrementation
- Added many "Trigger once" conditions

Game files to open the project (old)
[run-dino-run.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8530987/run-dino-run.zip)


4ian:
Nice! Super quick turn around on this :) 

- Use a custom font to make the score (and any text) better looking than the default Arial font.
- Score and Submit buttons are reversed it seems.
- We would need a menu with a title and a button "Leaderboard" and "Run"/Play.

Should we remember the name of the player between runs? (store it in a global variable?)

4ian:
And to highlight what I like:

- it's fast to play
- Gameplay is easy and fun.
- Perfect leaderboard integration.
- I love the explanation at the beginning. Cute and simple
- Graphics are fine
- Audio and sounds are fun!

arthuro555:
🦖🏃 

---

### General remarks

image
Missing yourself as an author

---

image
Missing game name in package name

---

image
I don't think any other example suffixes scene names with "Scene" 🤔 Not sure if we should remove them or not.

---

### Game over scene

image

- Cosmetics is a term that is often rather associated in casual games with hats, skins, and costumes. I'd rather use a term like "Visuals" or "Running dino" to make it clear what this events group is for
- Rather don't use the platformer behavior but a simple permanent force for moving the character constantly to the right, as that behavior can be missused in twisted ways by users that do not know any better and are not aware of better solutions if they are taught to use the behavior for every kind of movement.
- Do not put the dino to a run animation in events, rather, put it as such by default in the scene editor.
- The comment is not punctuated

---

image

Remove this, it is redundant and adds unnecessary complexity.

---

image

Lots of redundancy, use sub events (also always try to put objects conditions last for performance and replace the variable for the player name with using the expression for getting the inputed value directly): image

---

image

- Move this to a separate "Leaderboards" group, as they do change the buttons state as a side effect but are a separate concern from handling the buttons clicks
- Use trigger onces, changing the visibility or an effect takes a small performance hit when executed, doing it every frame is not very good. Generally, we want to give people the mindset of "if I only need this to happen once when that condition is true, I should always put trigger once by default" for performance reasons and also to make them avoid bugs in the future due to not using trigger once.
- image
This is a bug in the engine that we should fix in GDevelop, not through a workaround in the example.

---

### Game scene

*Already submitting because I have a bad history of github discarding what I had taken hours writing before I submitted it, and this post is already pretty long.*

4ian:
Thanks for the review! I confirm we should fix the input visibility in the engine itself.

Bouh:
I added the new thumbnail.

D8H:
Some nitpicks:
* Custom point can be defined on the platform to place the orbs.
* It's strange to change the Y right after the object creation. The cactus could have their origin point at their bottom center to avoid this.
* The background shift is missing a `TimeDelta`
* The events of DinoAnimation could be easier to understand if they were in "chronological" order.
* The usage of "simulate Up" and " simulate Down" is hacky. These controls are used for ladders and it introduces a 1 frame delay.
* Condition on `GameStarted` should be the root of any event to clearly too what happens when the game is started or not.
* There is no need to change a scene variable right before going to the GameOver scene as it will be destroyed anyway.

Bouh:
- For PC, Press SpaceBar to run.
- On mobile, we have to press which part of the screen to jump and slide?
- On mobile when the platform appears late for some reason, which doesn't happen in fullscreen:

https://user-images.githubusercontent.com/1670670/165073840-e4f108b0-a7f4-472b-a9c3-c19d45dcbdac.mp4

ClementPasteau:
tiny feedbacks on my end:

image
- Should we always try to use "released" instead of "down" -> this avoids clicking on something on a new scene by mistake because the scene switched and you had the click/touch down

image
- As you're already activating / deactivating hover at the bottom of the events based on the cursor position, the one at the beginning of the scene might not be needed?

Otherwise, it's really nice overall!

D8H:
Game:
* "GameStarted" and "GameRunning" are hard to follow. What about 3 states "Preparing", "Playing", "Dead" or something like this?
  * The trigger once is not necessary. With the "GameRunning" condition it won't repeat.
* Transition to "GameOver", the trigger once is not necessary.
* the background offset change should be linear to `TimeDelta()` if you want 1 pixel in most case, what about `60 * TimeDelta()`?
* "obstacle_spawn" can be rested in the same event as it's tested to be easier to follow.

DinoAnimation:
* The external events "DinoAnimation" do sound and input. What about "DinoMovement"?
* The "Ignore default controls" action is related to a property. It can be unchecked directly.
* The "animation of Dino is finished" condition doesn't need a trigger once as the animation is changed it won't stay finished.

ClementPasteau (approved):
LGTM! Should we merge it? :)

D8H:
> LGTM! Should we merge it? :)

I think we should wait the next automatic released. Leaderbord and touch functions will appear as unknown with the last stable release.

ClementPasteau:
good catch @D8H - let's plan the stable release then 👍

arthuro555:
I did a few additional changes:

---

Reordered the scenes to follow the game's scenes flow:

image

---

Modified this event to not use hardcoded values:

image

---

Removed this unused variable check:
image

---

Use objects groups:
image
⏬ 
image

---

Wrap the game states in groups:
image

---

Collapsed and colorized the group events as per the new guidelines (they were just updated)

---


I didn't touch it, but I want to bring to attention that the DinoMovement external event looks very hacky: seems like it relies upon the dive animation not being the same height for the character to be seen as not touching the floor but not jumping either to do the dive sound? That's not something we want to teach our users! Also, since it is only used there, there is no reason to make that an external event. There are already groups for organization, and this part is not really a separate module but an integral part of the system it is linked in, so it shouldn't be an external event.

Something we also probably want to do is to make the dino not global to not have to disable the platformer behavior on the other scenes and only have the animations needed for that scene loaded into each scene.

arthuro555 (changes requested):
See the last section of my last comment.

D8H:
> I did a few additional changes

Isn't groups supposed to be expanded?

ClementPasteau:
Merging this so we can make it public and start having creators use this good example 👍 
We can still bring a few changes if we feel it can be improved

PR #275, opened 2022-05-04
Add GDevelop account as author on game templates

Bouh:
Add the official GDevelop account on game templates.


4ian (approved):

ClementPasteau:
@Bouh you need to update the test file `exampleShortHeaders.json.spec.js` to change the authorId in there too :)

PR #284, opened 2022-05-16
Update virtual joystick example with multitouch joystick

tristanbob:
Includes three scenes:
-- Single joystick
-- Dual joysticks (twin stick shooter game)
-- Single joystick with buttons


4ian:
Had a quick look, seems to be working well, well commented, support for mouse/gamepads is here and the game logic is reduced to the minimum. I think it's good to go!

PR #285, opened 2022-05-17
Platformer updates based on user feedback (v1.2.1)

tristanbob:
Reported problems:

- When trying to add more objects to the level, users were confused by the hidden boundary objects which act like a hidden wall (due to platform behavior)

- Users have to manually update positions for camera limits when growing the level

Changes:

- Added Top/Bottom/Left/Right boundary objects
- Changed camera limits to use boundary objects (users can simply move these to change the size of the level)
- Added HelperObjects group and hide them at the beginning of scene (so users will see them in scene editor)

Other improvements:

- Stopped clouds from moving on Y
- Removed "Clouds" and "HelperObjects" layers
- Combined X and Y tween into single tween

## Playable game
https://games.gdevelop-app.com/game-5aef0c16-f8fb-40e7-b93c-09375a9257bb/index.html

## Screenshots

Screenshot 2022-05-17 103444
Screenshot 2022-05-17 103600

### DIFF
--- a/examples/platformer/platformer.json
+++ b/examples/platformer/platformer.json
@@ -16,9 +16,9 @@
     "scaleMode": "linear",
     "sizeOnStartupMode": "adaptWidth",
     "useExternalSourceFiles": false,
-    "version": "1.2.0",
+    "version": "1.2.1",
     "name": "Platformer",
-    "description": "",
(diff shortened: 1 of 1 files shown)


Bouh (approved):

Bouh:
LGTM

PR #290, opened 2022-05-19
Updates to Duck Game

tristanbob:
NOTE: The hitbox size has a huge effect on gameplay, so this needs to be play-tested thoroughly and adjusted if needed.

## Playable Version

(Updated to use original hitbox size)
https://games.gdevelop-app.com/game-251513a5-054e-430b-9a4d-9c1d81866b86/index.html

## Changes

First commit:
- Added the S key as a valid key to press when trying to exit a door.
- Deleted the random untitled external layout that wasn't doing anything.
- Went through each level and changed spike collision masks
- Removed unused assets.
- Removed an "always" condition.
- Added a few extra comments.

Second commit:
- Changed from "pause and change" to simply changing scenes
- Rename "DoubleJump" to "JumpsRemaining"
- Change all global variables to scene variable (except for "Level").
- Added a few TriggerOnce to optimize performance
- Remove particle emitter on level 1
- Remove blank effect from block object
- Increased height of top blocks on level 2 to prevent player from getting stuck after falling through bottom and pushing to the side
- Instead of using custom collision masks on hitbox object, resized hitbox object
- Increased size of duck hitbox so duck was not going far inside walls


4ian:
> NOTE: The hitbox size has a huge effect on gameplay, so this needs to be play-tested thoroughly and adjusted if needed.

- Correct, we might want to revert this. I could easily beat this level in the original version (first try):

  image

  But in the new version, the hitbox seems unforgiving and this feels very frustrating (while the original version was super fun, even if maybe too forgiving). 
  
  I think this is a general good thing to do in games: adjust the hitboxes/difficulty so that when things are "close", it's in the advantage of the player. In the case of this game, the smaller hitbox of the first version allowed to feel like you're super agile, while the new version gives a more "block" aspect to the duck.
  This gives the impression that we play against the game, rather than with it, and that the game is not "playing fair", because it requires you to do a complex double jump between walls, but gives you no room to do it.

  (I'm speaking about this jump, but also some others:
  image
  ) 

  Finally, i think the duck was not going too far into walls because it was made precisely for feet to stay outside :)
  image


- Also, it seems that a level is skipped in the version you shared (the 2nd level):
  image

HelperWesley:
We also don't consistently fit in to this hole anymore. 😅
image

tristanbob:
Thanks everyone for the great feedback!  I have addressed the problems and the game should now play the way it did originally, with the same hitbox size.

- Changed hitbox back to original size (20x24)
- Fixed logic that was skipping scene 2

BTW - I have updated the playable link at the top post to use this version.  I'll also include it in this post to make it easier to click:

https://games.gdevelop-app.com/game-251513a5-054e-430b-9a4d-9c1d81866b86/index.html

Bouh (approved):
Tested ✅ 

(Just to let you know, this game isn't mobile compatible)

Issue #291, opened 2022-05-23
Game Template - TheConvictionOfGunDude - Two separate versions - Desktop and Mobile


HelperWesley:
The game is a cross between the binding of issac and nuclear throne, with sprinkles of originality tossed in.

The game showcases how to use/create a shooting mechanic, upgrades, enemies with varying AI and attack patterns, leaderboards, health bars and other UI elements, multiple scenes, and external events.

The mobile version also demonstrates how to create dual joysticks using the multitouch joystick extension.

## Checklist

- [x] I've followed all of [the best practices](http://wiki.compilgames.net/doku.php/gdevelop5/community/guide-for-submitting-an-example).
- [x] My game has a proper name in the game properties. 
- [x] My game package name behind with `com.example.`.
- [x] My game has all events unfolded.
- [x] I've added myself as the author in the game properties.
- [ ] I've included a file called "README.md" with a description in proper English, explaining what this example is doing.
- [x] I confirm that this game and all of its resources can be integrated to this GitHub repository, distributed and MIT licensed.
- [x] I've cleaned unused resources.

## Game folder

[ConvictionOfGunDude_GithubVersion.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8758603/ConvictionOfGunDude_GithubVersion.zip)

[ConvictionOfGunDude_Mobile_GithubVersion.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8758605/ConvictionOfGunDude_Mobile_GithubVersion.zip)

HelperWesley:
Changed the Submit, reset, and submit score buttons so they highlight when hovered over on the desktop version.

This doesn't make sense on the mobile version, so it didn't get added to that version.

[ConvictionOfGunDude_GithubVersion.zip](https://github.com/GDevelopApp/GDevelop-examples/files/8767394/ConvictionOfGunDude_GithubVersion.zip)

HelperWesley:
![LiluoLogoMobileWithBanner](https://user-images.githubusercontent.com/72082749/170165484-24538d40-0307-4c2b-9a90-fb3bb5868d76.png)
![LiluoLogoWithBanner](https://user-images.githubusercontent.com/72082749/170165487-1ba1ef33-350c-4b4f-84e0-3128988d7ed9.png)


Thumbnails to display in engine on the homepage.

tristanbob:
Closing this issue as work will be continued in the two PRs listed above.

PR #293, opened 2022-05-25
The Conviction of Gun Dude (mobile version)

tristanbob:
- Initial import from #291


Bouh:
## General 
- [ ] Do you think it is possible to keep only one version?
Have a unique cross-play version like the Platformer game.
- [ ] Game description is missing
- [ ] Add the **GDevelop** account in the author
- [ ] Flag the `StartMenu` scene as first scene (done in desktop version, but no on mobile)

## Scene
- [ ] **Darkening** object can be deleted and replaced by the background color

## Events
- [ ] For **Reset_Button** object, and all other objects using actions to be responsive in the UI, please instead use the reworked Anchor behavior.
- [ ] I opened Desktop and Mobile versions, and the events are different, especially on `Start Game `group. This confirms we should keep only one version and make it cross-platform.

PR #315, opened 2022-06-14
Update README.md

Bouh:


README.md
```diff
@@ -22,11 +22,16 @@ If you've created an example with GDevelop, you can submit it to be shared with
...
-4. Submit it! You can either [submit it here](https://github.com/GDevelopApp/GDevelop-examples/issues/new/choose), attaching the _zip file_.
+3. Create a `preview.png` (case sensitive) 16:9 image for the thumbnail of the game example, and you can also add a `thumbnail.png` (case sensitive) with a 1:1 ratio to add a tiny thumbnail in the examples list of the game engine these assets are to add at the root of the game folder.
+4. Create a new `README.md` file and write a short description of the game.
+5. **Export** your game and all its resources to a zip file (you can save it in a new folder and zip this folder).
+6. Submit it! You can either [submit it here](https://github.com/GDevelopApp/GDevelop-examples/issues/new/choose), attaching the _zip file_.
 
 If you know how to create *Pull Requests*, you can also clone this repository and add your example in the examples folder, and then open a PR. Examples are deployed automatically when pushed to the `main` branch: [![CircleCI](https://circleci.com/gh/GDevelopApp/GDevelop-examples/tree/main.svg?style=svg)](https://circleci.com/gh/GDevelopApp/GDevelop-examples/tree/main)
 
+## Developper
```
Bouh, line 32:
```suggestion
## Developers
```

tristanbob (approved):
This is exactly the information I was looking for. 

You should add the min resolutions for the images.

I think it is 1280x720 for the preview but what size for the thumbnail?

Bouh:
Most of thumbnails are square, and use 64 or 128px, but for some if them
there is a 16:9 image, i think this need to be formated for the next
homepage.
And the definition or the preview and thumbail is unclear as we Can use one
or the other and it works same 🤷🏼‍♂️

Le mer. 6 juil. 2022 à 16:37, Tristan Rhodes ***@***.***> a
écrit :

PR #320, opened 2022-06-19
Absorbus: New game template

tristanbob:
## Playable game
https://liluo.io/victrisgames/absorbus

I'm not sure the best time to create a draft PR for a game template, but perhaps I'll make one now.  The game is 99% done, it just needs to be cleaned, polished, and play tested.

Here are some things to think about:
- How does the movement feel?  Should the player move faster/slower/different?
- Any ideas on how to make it juicier? 
- Better color selection (such as the button) or better UI design?  
- Different music or sound?
- Why does the end-of-level sound effect play at start of level? (bug)
- I loved the wobble effect, but Davy didn't.  I left it disabled because it causes hitboxes to not match anymore, and it also looked bad on the auto-zoom-out logic.
- Should we use the layer "texture" or delete it?
- Should we leave the extra images (from the same asset pack) and audio so players can start creating faster?
- There is a call to action at end of the game so people know they can make games like this
- How do we like the design of a single scene, lots of external layouts and external events? (I think it works well and makes it easy to modify/expand the game)


4ian:
Very nice, I love the progression with smart enemies. Would be great to publish on the Play Store too :) 
- The last link should link to https://gdevelop.io/game-example/absorbus
- This condition: image could be achieved with an inverted condition, not need I think for the "invert the logical result". Inverting a collision condition already checks for the player not colliding with ANY Enemy. (to be double checked in the game)
- There are a lot of disabled events in "Smart Enemy" external events: 
image
- The game overall is quite complex in terms of logic and could make a new user afraid of all the logic in external events. Not to do now, but think about how to make more extensions/custom behaviors to make the scene events as lean and clean as possible.

Approving this so you can merge once you've looked at these points :)

tristanbob:
> Very nice, I love the progression with smart enemies. Would be great to publish on the Play Store too :)

Thank you!  I agree it would be good to have on Google Play Store.  Can you create a new app so I can start that process?

> * The last link should link to https://gdevelop.io/game-example/absorbus

Fixed.

> * This condition: image could be achieved with an inverted condition, not need I think for the "invert the logical result". Inverting a collision condition already checks for the player not colliding with ANY Enemy. (to be double checked in the game)

Turns out that neither my logic nor yours was actually stopping the sound.  The "stop sound action" might need to be tested and confirmed but for this game, I simply removed that feature since the game seems fine with the full sound effect.

> * There are a lot of disabled events in "Smart Enemy" external events:

You found my "backups"!  I use this method for quick testing and rollback.  I have removed them.

> * The game overall is quite complex in terms of logic and could make a new user afraid of all the logic in external events. Not to do now, but think about how to make more extensions/custom behaviors to make the scene events as lean and clean as possible.

I agree that the event structure is complex.  I am trying to find the right balance of simplicity while eliminating duplicated events.  This game could help someone who is ready for a complex logic structure, and we can let the other game examples help beginners.

>   * Typically this. This could have been made into a custom behavior, or could be in the future, because we would need to allow custom behaviors to have **resources** as properties.

Yes, that sounds like a very useful thing to have!  Are we tracking that feature yet?
 
> Approving this so you can merge once you've looked at these points :)

I don't think you approved it yet.  

That said, I still need to add thumbnail/preview images before this can be merged.

4ian:
> Thank you! I agree it would be good to have on Google Play Store. Can you create a new app so I can start that process?

Done

> Yes, that sounds like a very useful thing to have! Are we tracking that feature yet?

Nope! Feel free to open an issue, as I think this needs some good thinking behind it.

> The "stop sound action" might need to be tested and confirmed but for this game, I simply removed that feature since the game seems fine with the full sound effect.

Strange! Anyway, no problem let's just remove this.

4ian (approved):
LGTM

PR #339, opened 2022-07-04
[Parking Jam] Remove the unused effect on layers

Bouh:
Remove the unused effect on layers


tristanbob (approved):
I can confirm that removing the "glow" layer effect drastically increased performance on my low-end Samsung Galaxy A11.  Thanks for finding this, @Bouh, now I can move forward with Google Play Store again!

PR #343, opened 2022-07-12
[Not-a-Vania] Add a collision layer.

D8H:
It needs this change in GDevelop:
* https://github.com/4ian/GDevelop/pull/3313
* https://github.com/4ian/GDevelop/pull/4106

* Project: https://www.dropbox.com/s/d1y00hezzswc0te/not-a-vania-collision-layer.zip?dl=1
* Build: https://liluo.io/games/c8389601-2960-4197-baa7-f92e3a976c00?dev=true

image


Silver-Streak:
I've done some testing, this seems great, however anywhere these objects appear, they behave differently. In my original project, you can run against these, hold right/left, and still jump up/over them. in your example project liluo link, you just wiggle in place and don't actually jump.


image

Bouh (approved):

PR #355, opened 2022-07-24
Absorbus 1.0.2: Added Pinch zoom (thanks @D8H!)

tristanbob:


D8H:
I only took a look to the events, I will check on my phone tomorrow.

- Where is the CopyCameraSettings extension used?
- Every other zoom action do it on 2 layers, should the CopyCameraSettings be used instead?
- Why use a "trigger once" instead of a "at the beginning of the scene"?
- Should "Zoom out as player grows" be disabled while the player is zooming to avoid it to fight against the changes? (maybe using a timer so it works with the wheel too)

tristanbob:
> * Where is the CopyCameraSettings extension used?

Looks like.... nowhere.  I think I was planning on using it but never did.

> * Every other zoom action do it on 2 layers, should the CopyCameraSettings be used instead?

I can't remember why, but I moved all objects off the "texture" layer you had made.  So the "texture" layer is not in use anymore.  I was reluctant to delete it entirely in case we wanted to roll back that decision.  

> * Why use a "trigger once" instead of a "at the beginning of the scene"?

Those conditions are basically the same, but I was concerned that the external event could be put behind a GameState variable condition, and then those events would never see "the beginning of the game".

> * Should "Zoom out as player grows" be disabled while the player is zooming to avoid it to fight against the changes? (maybe using a timer so it works with the wheel too)

Yes, it is "fighting" (pushing back) if you zoom in too far.  I don't consider this a major problem.  What would you like to see?

D8H:
> > * Every other zoom action do it on 2 layers, should the CopyCameraSettings be used instead?
> 
> I can't remember why, but I moved all objects off the "texture" layer you had made. So the "texture" layer is not in use anymore. I was reluctant to delete it entirely in case we wanted to roll back that decision.

I used this layer to be able to hide the textures and only show the colored disks because it's useful for debugging.

> > * Should "Zoom out as player grows" be disabled while the player is zooming to avoid it to fight against the changes? (maybe using a timer so it works with the wheel too)
> 
> Yes, it is "fighting" (pushing back) if you zoom in too far. I don't consider this a major problem. What would you like to see?

I guess it would be better to enforce a clean min and max zoom on user inputs.

Bouh (approved):

PR #363, opened 2022-07-29
AutoTiling Example

planktonfun:
## Describe the example

An example of early versions of autotiling in game

Autotiling lets you create tiles that automatically changes its frame based on its adjacent tiles.

another version of: [Floki](https://github.com/FlokiTV/GDevelop-AutoTile)

references: [Wiki](https://www.boristhebrave.com/2021/09/12/beyond-basic-autotiling/#:~:text=Autotiling%20is%20a%20system%20for,well%20supported%20by%20game%20engines.)

Sprite credits: [Cup Nooble](https://cupnooble.itch.io/)

## Checklist

- [x] I've followed all of [the best practices](http://wiki.compilgames.net/doku.php/gdevelop5/community/guide-for-submitting-an-example).
- [x] My game has a proper name in the game properties. 
- [x] My game package name behind with `com.example.`.
- [x] My game has all events unfolded.
- [x] I've added myself as the author in the game properties.
- [x] I've included a file called "README.md" with a description in proper English, explaining what this example is doing.
- [x] I confirm that this game and all of its resources can be integrated to this GitHub repository, distributed and MIT licensed.
- [x] I've cleaned unused resources.


D8H:
Hello thank you for sharing this.
Please can you attach a zip of your project?

D8H:
Looking at the tiles number, I guess you could simplify your implementation. Each edge can be a power of 2 (1, 2, 4, 8) their sum will give a value from 0 to 15 directly.

planktonfun:
here you go: 
[autotiling.zip](https://github.com/GDevelopApp/GDevelop-examples/files/9224150/autotiling.zip)

although you can test it directly through the web editor: https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/planktonfun/GDevelop-examples/examples/autotiling/examples/autotiling/autotiling.json

> Looking at the tiles number, I guess you could simplify your implementation. Each edge can be a power of 2 (1, 2, 4, 8) their sum will give a value from 0 to 15 directly.

Hmmm how do I do this?

D8H:
> here you go: [autotiling.zip](https://github.com/GDevelopApp/GDevelop-examples/files/9224150/autotiling.zip)
> 
> although you can test it directly through the web editor: https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/planktonfun/GDevelop-examples/examples/autotiling/examples/autotiling/autotiling.json

Thanks, good to know.

> > Looking at the tiles number, I guess you could simplify your implementation. Each edge can be a power of 2 (1, 2, 4, 8) their sum will give a value from 0 to 15 directly.
> 
> Hmmm how do I do this?

Something like this?

Tile16

What do you think of doing a behavior that watch instances creation and destruction with `onCreated` and `onDestroyed` (and maybe that check instances that was moved)?

planktonfun:
ohhh because the mapping looks like binary, yeah that works
- [x] binary sides

planktonfun:
> What do you think of doing a behavior that watch instances creation and destruction with `onCreated` and `onDestroyed` (and maybe that check instances that was moved)?

I tested this but there's a weird bug that oncreated its X and Y is always 0,0 even though you put it somewhere else in the scene

D8H:
> I tested this but there's a weird bug that oncreated its X and Y is always 0,0 even though you put it somewhere else in the scene 

The tiles can be updated in `onScenePostEvents`.

planktonfun:
> - [x] The tiles can be updated in `onScenePostEvents`.

Updated files

D8H:
Some suggestions:
- The extension should probably not destroy object instances created by extension users. They won't understand why.
- Define an action to update a tile instead of using variables to trigger something in another life-cycle function.
- Users might use rectangular tiles for a depth effect.
- Why not use a Map from a position to an object instead of 2 sets?
- `Set` can be iterated. Creating an array each frame is not efficient and should be avoided.
- By convention, functions start with a lower case in JavaScript.

planktonfun:
hmmm might be too advanced for rectangles we are not even using diagonals yet for an early version of auto tiling

D8H:
> > * Define an action to update a tile instead of using variables to trigger something in another life-cycle function.
> 
> Thats the first update but there's a bug with onscenepostevents where it cannot find its own extension functions

I though this bug was fixed. Which GDevelop version are you using?

> > * Users might use rectangular tiles for a depth effect.
> 
> Too advanced for early version of autotiling, we are not even using diagonals yet

This is to avoid to do a breaking change later. I guess there is not much to change to handle it.

> > * Why not use a Map from a position to an object instead of 2 sets?
> 
> because I can't search with `.has` using only one set

Your code is using `has` on only one of the sets.
I still think that a `Map<{x, y}, RuntimeObject>` will be better.

> > * `Set` can be iterated. Creating an array each frame is not efficient and should be avoided.
> > * By convention, functions start with a lower case in JavaScript.
> 
> There's a different ecma variation standards for javascript not sure what gdevelop is using, but it seems to change everytime per developer, feel free to change the code as you see fit.

If you use a `Set` which is ES6 you can iterate too. But if you need to check each element of a Set, it's probably that you don't need a Set or that you don't use it the good way.

You can use ES6 features:
https://github.com/4ian/GDevelop/blob/master/newIDE/docs/Supported-JavaScript-features-and-coding-style.md

PR #364, opened 2022-07-31
[platformer-with-tilemap] Add a linear camera following.

D8H:
Changes:
- Use a smooth camera instead of a direct centering
- Change the platformer character jump settings

### DIFF
--- a/examples/platformer-with-tilemap/platformer-with-tilemap.json
+++ b/examples/platformer-with-tilemap/platformer-with-tilemap.json
@@ -18,8 +18,8 @@
     "templateSlug": "",
     "useExternalSourceFiles": false,
     "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


D8H:
The camera can have a 1 frame delay if the platformer behavior is executed after it or if the object is moved by events. The SmoothCamera extension should probably let the user decide when to apply the following (between the object movement and the rending) to avoid this.

D8H:
* Camera following in the life-cycle function: 1 frame delay
https://games.gdevelop-app.com/game-c29f6b40-81cd-422f-b683-372675135e66/index.html

* Camera following after the player movement though an action: no delay
https://games.gdevelop-app.com/game-5d60504f-534f-4194-b9d0-b0b3649c3fbc/index.html

Bouh (approved):

PR #367, opened 2022-08-07
Ball Cup Boom (new game template)

tristanbob:
## Description

Sort 4 balls of the same color into a cup to make it explode. 
Complete all 10 levels to complete the game and climb the leaderboards!

This game was inspired by "Sort It 3D" and is a game template for GDevelop, a no-code, open-source, free, and easy game-making app.

## Play game

https://liluo.io/victrisgames/ball-cup-boom


4ian:
Just finished it! Don't forget to set a minimal score on leaderboards :)
I've not looked at how the game is made, but from the game perspective itself, it's super well made!

4ian (approved):

4ian:
I approved so this can be merged anytime you think this is good enough. 
I would love if you could move Explode Cup to an extension action as in theory that's "the way to do" when you want to do something reusable (we might blur more the distinction between external events and functions in the future).

tristanbob:
> I approved so this can be merged anytime you think this is good enough.

Thanks!

> I would love if you could move Explode Cup to an extension action as in theory that's "the way to do" when you want to do something reusable (we might blur more the distinction between external events and functions in the future).

The ExplodeCup external event has 4 purposes:

1) Trigger shake camera extension.  This won't work until extensions can have dependencies.

2) Create Particles. This could be done today by passing the particle object as a parameter.

3) Play 4 audio files.  This won't work until extension functions support audio resources as parameters.

4) Delete objects.  This could be done today by passing the particle object as a parameter.

We will have to wait for 1 and 3 to be available before we can migrate ExplodeCups to an extension, but I am looking forward to that day!

tristanbob:
> Just finished it! Don't forget to set a minimal score on leaderboards :) I've not looked at how the game is made, but from the game perspective itself, it's super well made!

I have added a minimum score and deleted my testing scores.  So far you have the best (and only) score on the leaderboard!  :)

And thanks for the compliment!

Bouh:
The change tab of this PR includes modifications of many games. Because of the master merge in this one.
Try starting a branch from master instead of merging, it's better for people doing a review.

PR #374, opened 2022-08-23
Health with shield (new example)

tristanbob:
190040503-64c0ae9e-7f1c-48dd-919c-543a2da11c09

## Playable game:

https://liluo.io/victrisgames/extension-health

## Summary:

Manage health (life) and shield points of an object. Health can be assigned to players, enemies, NPCs, or even inanimate objects such as breaking doors and mining rocks.

## How to use:

- Set desired values in "Health" and "Shield" behavior properties
- To apply damage use “Damage an object” or “Damage an object protected by a shield”
- To heal health, use “Heal object” action 
- To activate the shield, use “Activate Shield” action
- Health and shield regeneration will happen automatically based on properties
 
## Tips:

- While damage cooldown is active, no damage will be applied to health or shield
- Try using the “Blink” extension while damage cooldown is active
- Use the “Is just damaged” condition to trigger audio and visual effects

## Health properties:

- Health (life points) Used for starting health
- Maximum health (0 for no maximum)
- Damage cooldown (seconds) Use "0" for no cooldown
- Rate of health regeneration (points per second) Will not exceed max health, even with overhealing enabled
- Delay (seconds) before health regeneration starts after being hit
- Allow heal to increase health above max health (will not exceed max health)

## Shield properties:

- Max shield points (Shield will start with this many points when activated)
- Duration of shield (seconds) Use "0" to disable shield timer
- Rate of shield regeneration (points per second)
- Shield will block all excess damage when shield is destroyed
- Delay (seconds) before shield regeneration starts after being hit
- Regenerate shield when it has zero points


Bouh (approved):

HelperWesley:
Having a look through the extensions and these are the issues I found:

-I find 2 results for this when searching "damage".
image

-The healing for health and restoring/activation of a shield should work and be presented in the same way. Even if a shield didn't exist before and now does, and it isn't meant to take damage but just protect against a single hit, then giving a shield 1/2/3/etc points of health could be used to show how many hits it will take to break the shield.

image

The way it is now, the shield can't be restored for partial amounts, so if someone wanted to set up a system where 1 hit = 1 damage, and wanted multiple shields, they wouldn't be able to do that with this system.

Like this:
image

-Should % chance variables be 0-100 instead of 0-1? That reads clearer in my mind, since that's the way that information would be displayed to the player. 🤔
image

-After some testing with the "regen shield when it has zero points" option, I think it would be a strange mechanical choice to have a shield that regens until it's broken. I think people will want a hard shield or one that regens, but I can't think of any game off the top of my head that has regening shields that go away once broken. 🤔
image

tristanbob:
> -I find 2 results for this when searching "damage". image

This is a good catch!  One was for health and one for shield. Now they have descriptive words:

image

> The way it is now, the shield can't be restored for partial amounts, so if someone wanted to set up a system where 1 hit = 1 damage, and wanted multiple shields, they wouldn't be able to do that with this system.

I have modified the "activate shield" action so you can choose: 
- Shield points to the shield
- If shield duration will be renewed


> -Should % chance variables be 0-100 instead of 0-1? That reads clearer in my mind, since that's the way that information would be displayed to the player. 🤔 image

Most of the time GDevelop uses a range of 0 to 1 to represent percent, so I left it this way.

> -After some testing with the "regen shield when it has zero points" option, I think it would be a strange mechanical choice to have a shield that regens until it's broken. I think people will want a hard shield or one that regens, but I can't think of any game off the top of my head that has regening shields that go away once broken. 🤔 

Good point. I have removed this property and related functions.

tristanbob:
@Bouh @HelperWesley @D8H 

Any more thoughts on this?

The one thing I haven't tested is how well an old game works with the new health extension.  I don't think I renamed any properties or functions, and I only added parameters to existing functions.  Can someone do a test and confirm if this version is backward compatible with an existing game?

PR #377, opened 2022-08-27
Joint Connector Extension (example game)

tristanbob:
Things to look into:

- How will we get the required extension dependencies installed?

- Add actions/conditions/expressions so the joint values can be controlled (that is why I got the sliders and checkboxes prepared)

- The placement of overlapping objects and joint connectors is finicky.  I had to move everything around until the joints finally "worked".

- Sometimes the joints don't form properly on the initial ragdoll, but restarting the game usually fixes it.  In-game created ragdolls don't seem to have this problem.


tristanbob:
This PR has been replaced by #396

PR #387, opened 2022-09-24
Add an "Ellipse movement" extension example

D8H:
Show how to use the "Ellipse movement" extension to move a player in a snake-like game.

Extension PR:
- https://github.com/GDevelopApp/GDevelop-extensions/pull/591


tristanbob:
## Summary

This is a great example of a one-button game and you did it with only 2 events!  Bravo!

## Suggestions

- Use `CameraBorderLeft()` and similar expressions instead of `0` and `SceneWindowWidth`.  It's more obvious and it works even if the camera moves.
- Use `Object.Width() / 2` instead of `32` so people understand where the number came from (and it still works if they change the object image)

D8H:
>     * Use `CameraBorderLeft()` and similar expressions instead of `0` and `SceneWindowWidth`.  It's more obvious and it works even if the camera moves.
> 
>     * Use `Object.Width() / 2` instead of `32` so people understand where the number came from (and it still works if they change the object image)

Thanks, I did the changes.

PR #394, opened 2022-10-15
Added parameter inputs from the slider and checkboxes

planktonfun:
Added parameter inputs from the slider and checkboxes


tristanbob:
@planktonfun great progress! I was able to change the sliders and newly created ragdolls would use the new angle limits.  Very cool!  (Motors and breakable joints don't seem to be working, but I bet we can get that resolved.)

I noticed that you created many new parameters to use when creating the joints.  This works, but it is not the pattern that we like to use for advanced extensions.  Instead, try using the properties of the "Hinge (revolute) joint connector" behavior.  

To do this, we will first need to create expressions that return each of the behavior properties.  For instance, here is what the expression for MotorSpeed might look like:

image

Next, we need to enable the "JointOverlappingObject" function to access the functions we just created.  We can do this by adding a parameter that links the "Hinge (revolute) joint connector" behavior to the previous object parameter (so place this new parameter right below the JointConnector parameter).  This would look something like this (although I didn't move the parameter to the right place)

image

Did I make any sense?  Can you try these changes?  In summary:

1) In the "RevoluteJoint" behavior, create expressions that return each property of the behavior

2) Add a new parameter to "JointOverlappingObject" that links the "RevoluteJoint" behavior to the "JointConnector" object parameter. 

3) Change javascript to stop using parameter values as inputs, and instead use the new expressions that return the properties of the RevoluteJoint behavior.

Bouh:
To chime I had difficult knowing what has really changed in this PR.
There are unrelated commits:
image

tristanbob:
Yes I think so.  Did you push a commit with this update?

planktonfun:
@tristanbob not yet moving the PR

PR #397, opened 2022-10-19
Plinko game name changed

HelperWesley:
Didn't notice during submission, but the game folder name was "Plinkoexamplegame".

This change fixes that. Renaming the folder "plinko".


4ian:
Thank you! Could you also change the json name so that it's just "plinko.json"? Thanks!

4ian:
Nice! I did another change to avoid the capital letter to be 100% sure there is no problem between the folder name and the JSON name. Should be good now!

PR #398, opened 2022-10-21
Fire bullet example (gen3)

tristanbob:
This is used to test the new extension: https://github.com/GDevelopApp/GDevelop-extensions/pull/642


tristanbob:
The extension is finished so I'm ready for a review of this game example.

D8H (approved):
As it updates an existing example, I don't see any reason not to merge it, but this example contains a lot of unnecessary events. I think it should be made into a very simple example to avoid to confuse readers.

tristanbob:
> As it updates an existing example, I don't see any reason not to merge it, but this example contains a lot of unnecessary events. I think it should be made into a very simple example to avoid to confuse readers.

Thanks @D8H!  Are you talking about player and enemy movement? I want a minimal "game" experience that people could use to imagine this extension in their game.  (I removed perhaps ~90% of the original game code.)

D8H:
Oh, ok, the number of event group seems a lot but I didn't really look into them.

tristanbob:
> Oh, ok, the number of event groups seems a lot but I didn't really look into them. Maybe the 1st scene could be removed?

Next time I make a PR for this example (gen 4) I will remove the first scene.

PR #411, opened 2022-11-29
Resource bar - custom object example

D8H:


tristanbob:
Great work Davy! It blows me away that you created this in a day.

## Suggestions for Custom objects
- We need a description field for each child object to help explain what it is used for.
- Each child object should have a "Show this object" checkbox ("checked" as the default).  This way people can easily disable a resource and it won't show up as a magenta box.  In this example, the  "Show the label" property is hard to see and it would be more intuitive to be inside the matching Resource outlined area.
- Similarly, we should allow any parameter to be placed inside the outlined area of a resource. In this example, the "Animation" group of parameters should live inside the "Buffer/Delayed Bar" outlined area.

## Suggestions for "TiledContinuousBar"
- Change the child object named "Resource" -> "Filled bar"
- Change the child object named "Buffer" -> "Delayed bar"
- The label will always be overwritten by code, so we need a way to inform the users of this.  Is there any long-description field where we can add something like this:
```This label always shows the value of the bar using this format:  "Value / MaxValue". For example:  "50 / 100". ```
This could be located in the proposed "resource description field" I suggested above.
- doStepPostEvents condition is inverted.  It would be easier to read if the condition was NOT inverted and the direction of the inequality was changed (it seems the TiledUnitsBar already uses this method)
- I feel like I am missing some code.  Where is the logic that lerps the value of the "buffer"?

## PanelSpriteButton
- I tried to animate the size of the button using different "Default width/height" values.  "Hovered" was 10% bigger, "Pressed" was 10% smaller, and "Idle" was the normal size.  However, nothing changed when interacting with the object.  Should this method work?

image

D8H:
>     * doStepPostEvents condition is inverted.  It would be easier to read if the condition was NOT inverted and the direction of the inequality was changed (it seems the TiledUnitsBar already uses this method)
> 
>     * I feel like I am missing some code.  Where is the logic that lerps the value of the "buffer"?

This is a shortcut for "the timer value is greater than or doesn't exist".

> ## PanelSpriteButton
> 
>     * I tried to animate the size of the button using different "Default width/height" values.  "Hovered" was 10% bigger, "Pressed" was 10% smaller, and "Idle" was the normal size.  However, nothing changed when interacting with the object.  Should this method work?

I didn't intend it to work. I'm not sure we should make it works. I bet it won't look good as it won't zoom but just move the corners (and it would probably break the resizing system). It's a case we could handle with a sprite button.

Bouh:
## Bar

- [ ] There is two objects called "Resource bar", we should rename the first:
image
Into a **Progress Bar**.

To be able to differenciate it from the second which can kept the **Resource bar** name:
image

- [ ] the bar can't use decimal unit on the label?

- [ ] The buffer appear when we press minus, but its turned hidden when the plus is pressed. I played with a semi transparent background and it was weird to see something appear and disappear. I suggesto to keep it visible.
test

## Slider

- [ ] The thumb move when I release the mouse key, its not the common behavior, it should move when the mouse is pressed.
It make sense to use pressed condition, to avoid this weird things when we click on the thumb and release it. The feeling is better the the thumb follow the mouse in continue.
test

## Icon button  (probably in wip?)

Nothing to say I think its still in WIP? Because it dosen't work at all even with if I working on this version: https://github.com/4ian/GDevelop/pull/4643

## Toggle switch (probably in wip?)

- [ ] When the object have a custom size its completely broken.
- [ ] I Created a checkbox/switch from scratch and I can't display the label.
test
- [ ] The Default size of the checkbox isn't always the same by default (first text on two line, then text on one line, both is without custom size.)
test

D8H:
> * There is two objects called "Resource bar", we should rename the first into a **Progress Bar**. To be able to differentiate it from the second which can kept the **Resource bar** name.

"Progress bar" would be miss-leading. Both object can be used for the same thing. They are just displayed differently.

> * the bar can't use decimal unit on the label?

It would need a property to choose the number of decimals but I think it's not very common to have decimals on a bar so it rounds the value to keep it simple.

> * The buffer appear when we press minus, but its turned hidden when the plus is pressed. I played with a semi transparent background and it was weird to see something appear and disappear. I suggest to to keep it visible.

The buffer purpose is to catch the eyes because the value is important for players. Making the bar semi transparent would defeat the purpose. I think both case are mutually excluding.

>  * The thumb move when I release the mouse key, its not the common behavior, it should move when the mouse is pressed.
> It make sense to use pressed condition, to avoid this weird things when we click on the thumb and release it. The feeling is better the the thumb follow the mouse in continue.

Indeed, I fixed it.

> Icon button (probably in wip?)

It's just a test. I don't plan to release it.

> ## Toggle switch (probably in wip?)
> 
> * I Created a checkbox/switch from scratch and I can't display the label.

I plan to remove the label before the release if I don't find any good solution for it. When there is no child-instance visually encompassing any other, it's hard to decide how the children should adapt their size.

4ian:
Looks mostly good! A few things on my side:
* Is cheched => Is checked
  image
* This is useful, but a bit of a hack to show this initial value:  
  image
* "Thumb" as a name is very clear, but "Resource" is less. What about "Filling"?
  image
* The icon button is not showing at runtime but I guess it's not finished?

D8H:
>     * "Thumb" as a name is very clear, but "Resource" is less. What about "Filling"?
>       image

Indeed. Before I rename it, does anyone else have suggestions for the name?

>     * The icon button is not showing at runtime but I guess it's not finished?

It's just an idea that I forgave.

4ian:
@D8H: @HelperWesley tells that "fill bar" (so, "FillBar") would be the most understand next to "background" (and "thumb", for the slider).

PR #414, opened 2022-12-05
[Isometry] Simplify events

D8H:
* Use a dedicated extension to change animation according to movement direction.
* Fix collision masks for sound effects.
* Use the navigation mesh path finding extension for mouse control.


D8H:
> * Change navmesh shape-painter to be more visible.  Full opacity, color red, and width 2 pixels.  This is supposed to stick-out and be highly visible.

I think it's visible enough. If someone take a screenshot of it, at least it won't look too bad.

> * Change the animation based on the name instead of number. "idle" is easier to understand than animation "0".  (This does not apply to the movement function that requires the number)
> 
>     * Having two completely different "Is moving" conditions is very confusing (users have to observe the different icon and know the context of each condition)

Animation numbers are calculated, I moved it in the function to clarify.

> * If we are going to have footstep sounds in the grass, we need them for all types of terrain (at least a default sound for non-grass)

There is no foot steps sound in the resources of this example. Do you think the one from the platformer example could do?

> * "Move with Gamepad" is great and should be included in default controls of the top-down behavior (yes, the Gamepad extension would need to be installed by default).

I submitted an extension update.
- https://github.com/GDevelopApp/GDevelop-extensions/pull/690

Thanks for the review, I applied the other suggestions.

D8H:
I think there is no blocking issue. I'll merge it in a few days if nobody see something else to change.

PR #415, opened 2022-12-06
[Platformer] Extract events into functions.

D8H:
I won't do a changelog because I simplified and move event everywhere so the whole project must be reviewed anyway.

Unimplemented ideas
- The score could become a behavior on the text object, but maybe it would be going too far.
- Functions about the player could be moved in a behavior but it may cause 1-frame delay so I didn't do it.
- A button could be used for the game over (maybe we should wait to find a good one).


4ian:
Thanks for taking the time to improve this :)

Typos:
- Ennemy => Enemy (everywhere, name of extensions, behaviors and descriptions)
- MonterEnemmy => MonsterEnemy
- charcter => character

Other things:
- "Animate the character according to the platformer state." => "platformer state" is frightening in the description. 
- "BeKilled" is not a proper action name, because "Be" is not a proper verb here. It should be "SetKilled" or "SetDead" or "TriggerDeath". Prefer Trigger in this case.
- "_PARAM1_ fall into _PARAM3_" is not understandable. This is an action so there should be a verb, and fall is not understandable in this context. "Animate _PARAM1_ so that it's absorbed into _PARAM3_"
- Also rename the name and description ("Fall into the portal" => "Absorb into a portal")
- I don't like "Main" as the scene name, can it be "Level"? A bit more beginner friendly.

tristanbob:
I haven't tested mobile and gamepad controls yet, but everything seems to work so far (after adding the missing return value on a function).

tristanbob:
I tested this on a game controller and it works great.

Mobile controls work well, except that the jump button doesn't work the first time it is pressed.

Other than that, I think we have time to sneak one of those new custom buttons in to replace the "Retry" text.  :)

PR #425, opened 2023-01-02
[FireBullet] Added the ability to specify the layer that bullets are created on

tristanbob:
And removed starting scene because it is not related to the extension.


D8H:
The build fails because of a bug in the build script. It's fixed but it needs branches to merge master or rebase on it.
```
TypeError: gd.UsedExtensionsFinder.scanProject(...).toNewVectorString is not a function
```

tristanbob:
> The build fails because of a bug in the build script. It's fixed but it needs branches to merge master or rebase on it.
> 
> ```
> TypeError: gd.UsedExtensionsFinder.scanProject(...).toNewVectorString is not a function
> ```

Thanks for the advice; hopefully I did it right. It did pass the tests so that's a good sign.  ;)

PR #434, opened 2023-01-16
[Game feel] Refresh

D8H:
It uses  screen shake, smooth camera and fire bullet. It makes the example a bit easier to follow.

We should check if other extensions can be used or extracted, but this can wait another PR.


4ian (approved):
Good improvements already, merging this as it's important as this example is shown a lot.

PR #435, opened 2023-01-17
Extension example: Follow multiple objects with camera

tristanbob:
@D8H here is my (partially cleaned up) example for the extension update here: https://github.com/GDevelopApp/GDevelop-extensions/pull/710


D8H:
It's a great example, I have some suggestions:
- Events could forbid to remove the last ship. It will allow to remove the borders, the cat and the coins.
- "Make background follow base layer", it should be right under the extension action as it's about camera.
- "Save slider values as variables", is it to see them in the debugger? Why to use the extension value directly in the text actions?
- The "UI" group should probably come last. Even if the ships movement is not the point of this example, it may interest users.
- The slider max value could used to avoid the multiplication.

tristanbob:
> * Events could forbid to remove the last ship. It will allow to remove the borders, the cat and the coins.

I'm not sure what you mean by this.  Can you explain further?

> * "Make background follow base layer", it should be right under the extension action as it's about camera.

Done, I even changed to use the Copy Camera Settings extension.  :)

> * "Save slider values as variables", is it to see them in the debugger? Why to use the extension value directly in the text actions?

Done, changed to using expressions from the sliders.

> * The "UI" group should probably come last. Even if the ships movement is not the point of this example, it may interest users.

Done.

> * The slider max value could used to avoid the multiplication.

Done (this example was created long before you added that feature to sliders)  :)

PR #440, opened 2023-01-24
[Top-down RPG] Add multitouch controls

D8H:
### Changes
- Multitouch controls to move in the word map
- Use custom object for bars
- Update the button extension
- Better FSM
- Use a resolution 960x540 to use custom objects from the asset-store more easily.


4ian:
> Use a resolution 960x540 to use custom objects from the asset-store more easily.

Why was this useful?

4ian:
Looks much better with touch controls, thanks!

I just saw that you deactivated the condition/action to hide the touch controls, I'll let you change this if you want.

D8H:
> Looks much better with touch controls, thanks!
> 
> I just saw that you deactivated the condition/action to hide the touch controls, I'll let you change this if you want.

Yes, that was for testing, I will enable it back

PR #441, opened 2023-01-25
[Dino run dino] Use button

D8H:
* Fix a glitch when the dino dies under a platform.


4ian:
Seems like a resource is not used?

image

PR #446, opened 2023-01-26
[Tappy plane] Simplify events and comments

D8H:


4ian:
We should probably remove the player input in favor of the builtin authentication (open the banner at the game opening, like in the onboarding tutorial?).

D8H:
I'll add the banner after the next GDevelop release because the JS function is not exported.
It still handle player login if they already are connected.

PR #453, opened 2023-01-30
New example: Object Slicer

tristanbob:
Test the Object Slicer extension, which can slice an object into smaller pieces that match the color of original object.  

Useful for creating death animations such as explosions, crumbling, dissolve, or teleportation effects.
Pieces should be a solid white color or the color of the piece not match the original color.

Tips:
- Pieces are linked to the original object which can be used in other events (i.e. explode away from the original object) 

## Playable game
https://liluo.io/victrisgames/object-slicer


HelperWesley:
It's a really cool effect, but I got it to completely freeze up by cranking it up to 32x32 and pressing the explode button.
(Though that's not the act of splitting the image up, that's the resulting explosion causing that)

tristanbob:
I should have made a version that didn't use the physics behavior to see if it is slower. 

BTW, 32x32 is creating 1024 new objects per original object.

PR #456, opened 2023-02-01
Sokoban and papa-is-you example games

HelperWesley:
These example games show users how to create a sokoban game first, and then to create win conditions, and then finally how to merge and create strings to change logic in their games.

The games teach:
-Use of every type of variable
-Object positioning
-Object picking
-String manipulation and usage
-Raycasting
-Arrays
-Resetting a scene


D8H:
I only read the Sokoban example. It's very well written, good job.

Some suggestions
* The name "TempToggleVariable" doesn't allow to understand what the variable means.
* The marker could move using its own position so we understand it moves.
* The reset doesn't need any condition (the condition cost more than changing the value).
* The events from "Direction Check Markers" can be sub-events of its 1st one (and eventually "Win Game" too).

What do you think about using tweens to move?
Does it makes the events look too overwhelming?

I think one PR per example would have been easier to follow and one example may be ready before the other.

HelperWesley:
Thanks.

-I'll change the boolean variable to "LockControlsToggle", so it's easier to understand.
-I'll change the marker to move based on it's own position.
-I'll removed the condition for resetting variables. (I didn't know conditions were more costly than the actions)

As for tweens.... I think it would look better.... But maybe I'll leave the sokoban example as it is and test tweens on the "papa is you" example. Since that one is already going to be complicated, 1 more layer of complexity won't hurt.

If it makes it easier, I can submit 2 different PRs, so we can get the first example through first and then work on the second.

Let me know. 👍

Bouh:
Sokoban game is now added!
https://github.com/GDevelopApp/GDevelop-examples/pull/458

PR #459, opened 2023-02-03
Papa Is You Game Example

HelperWesley:
Building upon the sokoban example gamein GDevelop, this example uses mechanics similar to those found in Baba is You, a hit indie puzzle game made by Arvi Teikari in 2019.

The games teach:
-Use of every type of variable
-Object positioning
-Object picking
-String manipulation and usage
-Raycasting
-Arrays
-Resetting a scene


D8H:
Cool example

I have some suggestions:
* Pushing checks only need to be done if the player moved and rule changes when something is pushed.
* `LockControlsToggle` doesn't allow to understand what it means when it's `true` or `false`.
  What about "IsAnyDirectionChosen" or "IsAnyArrowPressed"?
* Why did you use ray cast? Was there any issue with the "point is inside" condition?
* Some code is copy-pasted. `HorizontalText` and `VerticalText` should be split in 2 variables one for the subject and another one for the verb. It will allow to do a helper function that set the right variable and functions to check the 4 variables.
* Why do you use 2 markers in this example?

HelperWesley:
Thanks for looking it over. 👍

-We would need another variable, or something, to let the event know if those things happened. But yeah, that can be changed. 👍

-The word toggle was to explain how it works, but I agree it's not as descriptive as it could be. It could be changed, but wording comes from this wiki page: https://wiki.gdevelop.io/gdevelop5/tutorials/how-to-make-togglable-states-with-variables

-Raycasting was just the first method that came to mind. Is point inside less demanding?

-I don't understand how a 4 variables and 2 sets of conditions would make that part better. I can't visualize it and don't exactly know what you mean by "helper function".

-In the first example, I used 1 marker, because once it collides with a wall we can just pick all objects that are tagged to move, and change their object variables. But in this case when the marker hits a wall, we need to remove the tag on a specific stack of objects. Since it needs to be done without effecting the positioning and movement of any other object that's not in that stack. So I created a second marker to go backwards along the stack and reverse what the original marker had done. We could change it to using a single marker, and give that marker an object variable to tell it when to change an object it's in collision with to "move" or "none", but that's more conditions and actions in the events to achieve the same thing.

I'll be working on a video tonight, but I'll fix up the issues and check back with this later this week. 👍

D8H:
With the tween animation, the trigger once on key press no longer work (because of the extra condition) so keeping a key press will make several moves.
It would need a new variable to ensure the key was released at some point. Do you think we should add it?

D8H (approved):

examples/papa-is-you/README.md
```diff
@@ -0,0 +1,3 @@
+Building upon the sokoban example gamein GDevelop, this example uses mechanics similar to those found in Baba is You, a hit indie puzzle game made by Arvi Teikari in 2019.
```
D8H, line 1:
```suggestion
Building upon the sokoban example game in GDevelop, this example uses mechanics similar to those found in Baba is You, a hit indie puzzle game made by Arvi Teikari in 2019.
```

examples/papa-is-you/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+In this example you play using typical sokoban controls, but you'll also move text blocks that can turn on and off different logic to solve puzzles in the game. With the eventual goal of each level being to put an object with the "win" logic in collision with an object with the "controlled" logic.
```
D8H, line 3:
```suggestion
In this example, you play using typical Sokoban controls, but there is a twist: you'll also push text blocks that change the game rules to solve puzzles.
```

PR #466, opened 2023-02-10
[FireBullet] Added bullet speed variance

tristanbob:
See: https://github.com/GDevelopApp/GDevelop-extensions/pull/778


HelperWesley:
I like having this extra variable, it means you can do short "fluid" bursts from enemies that move at different speeds. 👍

PR #472, opened 2023-02-14
Add the example size attribute to headers

D8H:


scripts/generate-database.js
```diff
@@ -259,6 +259,19 @@ const extractExamples = async (
 
         const project = loadSerializedProject(gd, projectObject);
 
+        /** @type {number} */
+        const eventsSize = gd.InstructionsCountEvaluator.scanProject(project);
+        /** @type {string} */
+        const eventsSizeLevel =
+          eventsSize < 20
+            ? 'tiny'
+            : eventsSize < 100
```
4ian, line 268: I would go for 70 for "small".

scripts/types.d.ts
```diff
@@ -33,6 +34,7 @@ export interface ExampleShortHeader {
   authorIds: string[];
   previewImageUrls: string[];
   gdevelopVersion: string;
+  eventsSizeLevel: string;
```
4ian, line 37: eventsCountLevel maybe? I'll let you choose.

4ian (approved):

scripts/types.d.ts
D8H: I'm thinking about `codeSizeLevel` because I don't want it be mistaked for a number.

PR #476, opened 2023-02-23
Add more information about the extensions of examples

4ian:


D8H (approved):
It seems to add anything we need for the example pages.

PR #488, opened 2023-03-19
[City builder] Handle touch devices

D8H:
### Changes
- Buildings can be dragged an dropped.
- Selected building can be unselected by clicking a 2nd time.
- Add extension for the toolbar and floating texts.
- Remove some useless trigger once


4ian:
Tested on my phone, works perfectly. Simple and nice example. I like the two extensions too, simple and useful.

PR #500, opened 2023-03-26
[Inventory] Rewrite the extension with events

D8H:


Bouh (approved):
LGTM for a merge.

D8H:
Sorry, I forgot to make it a draft.
The save and load action doesn't work because the extension variable has a dot. It could be fixed easily, but I wonder if we should wait to have variable parameters to give the exact same features as the original extension.

PR #506, opened 2023-04-03
New example for Advanced Projectile extension

tristanbob:
## Advanced Projectile
- Control how a projectile object moves including speed, acceleration, distance, and lifetime.

## Playable game
https://gd.games/victrisgames/extension-advanced-projectile

## Video
https://user-images.githubusercontent.com/8879811/229410580-9b90add5-2602-46f7-989b-63f16fa57a32.mp4


HelperWesley:
I like the extension because I can immediately see how I would use it in a game.

The deal breaker for me is what happens when you change a variable. The change is applied to all existing bullets, but I think it should just apply the change to newly created projectiles, not existing ones. If the player or enemy is switching between attacks, this could cause existing bullets to act erratically.

PR #523, opened 2023-05-14
Test fire bullet update

D8H:


tristanbob:
I tested this game and it worked great!  Thanks for adding this feature.

PR #525, opened 2023-05-16
Add 3D Shooting gallery example by HelperWesley

AlexandreSi:


D8H:
The multiple boolean conditions could be replaced by a small FSM though I don't think it's mandatory because the example is small and people will probably use it check how the camera is handled.

PR #546, opened 2023-07-17
Adding 3D lane runner game example

HelperWesley:
This 3D lane runner game shows the basics of adjusting a 3D camera, variables, sound effects, saving and loading a highscore, and how to use the platformer behavior in a 3D space.


D8H:
I think that unused 3D model animations should not be removed. Unused animations are a good way to generate a will to build over the example.
I didn't replace the character animations but for a next example, it would be better to keep them.

D8H:
How much air control should there be?

Currently players can move left and right:
- from an empty square
- in the air

D8H (approved):

PR #554, opened 2023-08-05
[Flash Object] Learn how to use the "Flash object" extension

tristanbob:
### Problem (Update: Solved!)

@D8H can you take a peek at this game and extension?
My problem might be caused from copy/paste functions from the FlashColor extension into the Flash extension, or perhaps renaming the behaviors.

This extension works for the first object "Red Hero".  
But when I add the behavior to a second object "Cyberpunk", it doesn't work.
Here is the strange thing: The objects have different behavior names!

RedHero has "FlashOpacityTween" behavior and it works (this was a previous name of the extension)
Cyberpunk has "FlashOpacity" behavior and it does NOT work

The behavior is currently named "FlashOpacity".
How is RedHero even working?
How do I get Cyberpunk to work?


tristanbob:
Hooray!  I fixed this by:

- Removing the old behavior name in `behaviorsSharedData`
- Search replace game JSON from "FlashOpacityTween" to "FlashOpacity"

D8H:
> Hooray! I fixed this by editing the game JSON file:
> 
>     * Removing the old behavior name in `behaviorsSharedData`
> 
>     * Search replace game JSON from "FlashOpacityTween" to "FlashOpacity"

Ok, I'll try and see if I can find a bug in the rename refactoring.

tristanbob:
> > Hooray! I fixed this by editing the game JSON file:
> 
> > 
> 
> >     * Removing the old behavior name in `behaviorsSharedData`
> 
> > 
> 
> >     * Search replace game JSON from "FlashOpacityTween" to "FlashOpacity"
> 
> 
> 
> Ok, I'll try and see if I can find a bug in the rename refactoring.

Don't spend much time on this; I could have broken it with a manual file edit.

PR #571, opened 2023-09-09
[Button effects] New example

tristanbob:
## Description

Change the way a button looks based on the state of the button

This is a simple yet powerful method for changing the appearance of objects using the "Button effects" behavior. 
The states of a button include: idle, hover, pressed, held outside, and focused.

Multiple behaviors are provided and they can be combined to make a juicy button with animated responses to user input.

- Button effect
- Button animation (number)
- Button animation (name)
- Button size (scale) tween
- Button color (tint) tween
- Button opacity tween
- Button angle tween

## Playable game

https://gd.games/victrisgames/button-effects

## Video

https://github.com/GDevelopApp/GDevelop-examples/assets/8879811/d059877f-cb9a-4c55-9c89-bfbff178fbbe


D8H:
The script doesn't handle spaces in project file name.
* https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/GDevelopApp/GDevelop-examples/button-effects/examples/button-effects/Button%20Effects.json

tristanbob:
> The script doesn't handle spaces in project file name.
> 
> * https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/GDevelopApp/GDevelop-examples/button-effects/examples/button-effects/Button%20Effects.json

Thanks, I have updated the filename to not use a space.

D8H:
@tristanbob If you're ok with the changes, I'll generate the getters and setters.

As discussed, I'll remove the opacity behavior.
What do you think about removing the animation by index behavior too?
We should probably postpone the angle behavior because of the collision mask issue to take more time to find a good solution. What do you think?

### Example changes
- Button are setup to have a fade-out 2 time longer than their fade-in

Other than this the example should work exactly the same.

### Extension changes

#### Features
- Add a behavior to tween effects.
- Remove Hovered and HeldOutside states handling as they are part of the Focused state.
- Duplicate easing and duration property to allow different speed for fade-in and fade-out.

#### Good practices
- Replace "trigger once" with a previous state check as "trigger once" are shared by all objects.
- Fix tween name collision between behaviors.
- Use the same tween name for every state because only one must play at the same time.
- Use default behavior names as required property behavior names to allow copy-paste of events between extensions and layout events without any issue.
- Slightly reword behaviors description

tristanbob:
> @tristanbob If you're ok with the changes, I'll generate the getters and setters.

Thanks, please do.

> As discussed, I'll remove the opacity behavior. 

Yes, this is fine.

> What do you think about removing the animation by index behavior too? 

I don't want to remove it.

> We should probably postpone the angle behavior because of the collision mask issue to take more time to find a good solution. What do you think?

I think we should ship the angle behavior as is. In many cases, it works great and we can always improve it later.

> ### Example changes
> * Button are setup to have a fade-out 2 times longer than their fade-in

This is fine and looks good.

> ### Extension changes
> #### Features
> * Add a behavior to tween effects.

This is a powerful feature, thanks.

> * Remove Hovered and HeldOutside states handling as they are part of the Focused state.

This is good.  BTW, found a spelling error in extension: "focussed"

> * Duplicate easing and duration property to allow different speed for fade-in and fade-out.

Does "fade-in duration" make sense as a name? I guess it does; the change is fading between two states. (At first I thought it only applied to opacity)

> #### Good practices
> * Replace "trigger once" with a previous state check as "trigger once" are shared by all objects.

I was wondering about this.  I'm glad you identified the issue and fix!
 
> * Fix tween name collision between behaviors.

Great fix.

> * Use the same tween name for every state because only one must play at the same time.

I like this change.

> * Use default behavior names as required property behavior names to allow copy-paste of events between extensions and layout events without any issue.

Why do we force users to name the required behaviors property?  I think the property names for required behaviors should be set automatically.

> * Slightly reword behaviors description

Looks great.

I also noticed you added a "TimeDelta()" expression that applied the TimeScale for the layer.   This is interesting because I use TimeDelta in several extensions and I just assumed it would change automatically with TimeScale.

Thanks for spending your valuable time reviewing and improving this extension!

D8H:
Should this example be merged with this one?
https://gdevelop.io/game-example/free/buttons

PR #582, opened 2023-10-15
[First person] Add raycast usage

D8H:


D8H:
> This raycast condition should be an object-less condition i think.

By "object-less" do you mean it should be in "Other conditions" or that it should not have any object parameter?

Bouh:
In "Other conditions"

D8H:
Why this condition and not the other?
I mean they all do the same thing, the only difference is how the ray is described.

PR #595, opened 2023-11-15
Ghost Racer (featuring new "Record Movements" extension)

tristanbob:
"Ghost Racer" is similar to a 2D Fall Guys game with a completely rewritten RecordMovements extension that will allow asynchronous multiplayer by playing against the movements of yourself or others on the leaderboard!

**Recording:**
Start/stop/pause recording
Select which properties to record
Set max recording duration (default: 60 seconds)
Set recording frequency (default: 20 FPS)
Set player name (if not set, will try to use Authenticated user, and fallback to use the object name)
Set score (used for sorting best recordings)
Save to Firebase (compression enabled by default, but optional)
Save to local storage (compression enabled by default, but optional)

**Playback:**
Create objects using shortest recordings (Firebase or local storage)
Create objects using highest scores (Firebase or local storage)
Start/stop/pause playback
Select which properties to playback
Interpolation (enable/disable)
Set playback speed, reverse direction
Move playback frame by a relative amount or go to a specific value


tristanbob:
I've cleaned this up a lot, but it still needs methods for recording/playback to local storage.

PR #603, opened 2023-12-03
[Flash object] Added functions to get/set the half-time, which controls the speed of flashing

tristanbob:


Bouh:
Object capabilities changed somes event that need to be updated by selecting them a new time in the instruction editor.
image

tristanbob:
> Object capabilities changed somes events that need to be updated by selecting them a new time in the instruction editor. !

Woah! I just discovered that "show deprecated items" option in preferences!  I think it should be enabled by default!

I'm updating this now.

PR #650, opened 2024-06-17
Paddle Battle: A simple multiplayer game based on Pong

tristanbob:
Based on the classic game Pong, Paddle Battle is a simple multiplayer game where two players compete to bounce a ball off their paddle.

Here is the playable link to Paddle Battle:
https://gd.games/VictrisGames/paddle-battle


ClementPasteau:
👀

PR #658, opened 2024-06-27
[Arrow Fight] [Example] Update Arrow Fight to support gamepads, touchscreen, and WASD controls

tristanbob:
Game
- Added player username (if authenticated)
- Stick dead players to arrows
- Added black outline to arrows

Controls
- Add gamepad controls
- Add virtual joysticks (twin stick) controls
- Added player state to help reduce control complexity

Playable game:
https://games.gdevelop-app.com/game-499fa3e3-233f-421c-8944-2a7ca5ad26cf/index.html


tristanbob:
TODO: Solve this bug - When player A shoots an arrow while touching the stopped arrow of player B, player A dies.

@HelperWesley

tristanbob:
> Can we use a loop to reduce this spagetti, and make it scalable? Something like, For each Score, and a local variable as index to loop on all players. 

I like your idea a lot (using a local variable with a repeat event) but I couldn't find a way to specify the "Wins.1" structure variable with the local variable.

tristanbob:
> TODO: Solve this bug - When player A shoots an arrow while touching the stopped arrow of player B, player A dies.
> 
> @HelperWesley

I solved this!  This is now ready for review and testing.  Play game here:

https://games.gdevelop-app.com/game-518be14c-c2e3-49ac-8926-79d4dd529710/index.html

tristanbob:
> > Can we use a loop to reduce this spagetti, and make it scalable? Something like, For each Score, and a local variable as index to loop on all players. 
> 
> 
> 
> I like your idea a lot (using a local variable with a repeat event) but I couldn't find a way to specify the "Wins.1" structure variable with the local variable. 

Can I use something like "Wins.LocalVariable(ThisPlayerNumber)"?

tristanbob:
I just realized that this logic is not right.  The "Player is Authenticated" describes the person running the game, not the player who earned each score.

I guess I  could change the condition to `Multiplayer::PlayerUsername(PlayerNumberIndex) != ""`?

tristanbob:
Yes, this should work better.  

@ClementPasteau can we get a new condition added for `Player username of player _integer_ is _string_`?

image

Test this game: https://games.gdevelop-app.com/game-3613b36f-9a12-485d-a8e1-5b35583718a2/index.html

tristanbob:
@Bouh are you happy with theee changes? Want to test the game?

ClementPasteau:
> @ClementPasteau can we get a new condition added for `Player username of player _integer_ is _string_`?

Is this to test if the username is defined or not? I'm not sure comparing a username with a string is very useful.
We could instead:
- automatically return `P2` `P3`... when the username doesn't exist
- or give a condition `player X has a username`

tristanbob:
You might be right; in this example I only need to know if there is a username, so your second option would work.

Your first option is not great because we don't know what the creator wants to use as an alternate string.

I'm trying to come up with reasons for a string comparison; perhaps to compare entries on a leaderboard?

PR #697, opened 2024-09-12
[Shark frenzy] a game inspired from Hungry Shark

D8H:


D8H:
### Changes
- Use anchor behaviors to center menu items.
- Add spaces around operators and after commas in expressions.
- Fix the expressions that uses `TimeDelta()`.
  - At 60 fps `TimeDelta()` is around `1/60` so, to keep the same behavior, you need to multiply the factor you chose (when `TimeDelta()` wasn't used) by 60.
- Replace `round(Multiplier)` by `floor(Multiplier)`
  - Otherwise, players need to gain 0.5 to reach 2 and then 1 to reach 3, 4...
  - It makes the game more difficult, so the multiplier should probably increase faster to compensate.
- Simplify the boost logic by using the max speed as the state.

PR #710, opened 2024-09-28
Advanced shape based painter example

ddabrahim:
Updated the example to comply with latest guidelines

Replaced number variables with boolean variables where it was reasonable to do so
Made sure all variables declared in the variables editor
Added extra comments and events groups to make it more simple to follow how the example works
Added RGBA letters to color selection lines.

### DIFF
--- a/examples/advanced-shape-based-painter/advanced-shape-based-painter.json
+++ b/examples/advanced-shape-based-painter/advanced-shape-based-painter.json
@@ -1,13 +1,15 @@
 {
   "firstLayout": "",
   "gdVersion": {
-    "build": 99,
-    "major": 4,
-    "minor": 0,
+    "build": 213,
+    "major": 5,
+    "minor": 4,
     "revision": 0
   },
   "properties": {
     "adaptGameResolutionAtRuntime": false,
+    "antialiasingMode": "MSAA",
+    "antialisingEnabledOnMobile": false,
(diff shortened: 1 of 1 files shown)


ddabrahim:
Preview

[https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/ddabrahim/GDevelop-examples/Advanced-shape-based-painter/examples/advanced-shape-based-painter/advanced-shape-based-painter.json](url)

ddabrahim:
It seems if previewed inside the web editor, the RGB color selection does not work. But if previewed  inside the desktop version and if I exported to local HTML5, it works. Not sure why it doesn't work inside the web editor.

ddabrahim:
I think I've found the problem. It seems the project is corrupted and the fact it is working on the desktop is pure luck. I managed to find a fix in the web editor, I am going to attempt to implement the fix on the desktop now and see if it also fixes it in the web editor

ddabrahim:
I can confirm that the project is fixed and working as expected in both the web editor and the desktop editor now. 
What I have done was:

Variable types in the variable editor was wrong
Replaced old variable events with the new variable events
Some variable expressions were also wrong, was calling a number variable when it was a string variable (RGBA).
Changed object names to remove errors regarding being the same as variable names.

 I am uncertain why is the build test faile however.

PR #715, opened 2024-10-07
Remove deprecated examples

D8H:
Also update a few extensions in some examples.


4ian (approved):
Thank you, that will be better for everyone to not have these as they have become mostly anti pattern thanks to improvement in extensions/in GDevelop :)

PR #724, opened 2024-11-05
Add screen orientation checker and improve title screen centering for 3D Car Coin Hunt

4ian:


examples/3d-car-coin-hunt/3d-car-coin-hunt.json
```diff
@@ -4276,6 +4396,943 @@
...
+          "functionType": "Action",
+          "name": "onFirstSceneLoaded",
+          "sentence": "",
+          "events": [
+            {
+              "type": "BuiltinCommonInstructions::JsCode",
+              "inlineCode": [
+                "const orientation = runtimeScene.getGame().getGameData().properties.orientation;",
+                "",
+                "eventsFunctionContext.globalVariablesForExtension.get(\"TargetOrientation\").setString(orientation);"
```
D8H, line 4439: Why not define an private expression for it?

examples/3d-car-coin-hunt/3d-car-coin-hunt.json
```diff
@@ -4276,6 +4396,943 @@
...
+                  "type": "BuiltinCommonInstructions::Standard",
+                  "conditions": [],
+                  "actions": [
+                    {
+                      "type": {
+                        "value": "ResetObjectTimer"
+                      },
+                      "parameters": [
+                        "Object",
+                        "\"Animate\""
```
D8H, line 4924: I think it's better to use children. I'm not sure the parent object will still work the same in the future.

Also, the timer is probably unnecessary as tweens has "is finished" condition, so you can loop between 2 tweens directly.

4ian: Done with "is finished", but I also had to remove tweens, it was painful to find how to do it.

examples/3d-car-coin-hunt/3d-car-coin-hunt.json
4ian: I'm thinking that in the future we could have an action to force a specific orientation - to override what is in the game (because some games might want to change orientation according to the scene... even if unlikely).

PR #763, opened 2025-02-10
Use Quick join lobby action in Multiplayer Jump and Arrow Fight games

AlexandreSi:
To merge when 5.5.225 is used everywhere


4ian (approved):

ClementPasteau (approved):
nice 🚀

PR #960, opened 2026-02-27
Animated score counter variant update

HelperWesley:
An update to the score counter object with new variants of the object, and variants that use the bitmap text object instead of the default text object.

Along with the new variants, there were some changes made to the extension in order to add the bitmap text, and fix an issue with the object. Those changes are:

1. I added a bitmap text object to the default variant and put it in a group called "CounterText" with the normal text object. Then I replaced all of the expressions/conditions/actions inside of the extension that used the normal text object "Counter" with the "CounterText" object group.

image

2. I then set the action to start shaking the text object to only use the normal text object "Counter", so that the bitmap text wouldn't be shaken when the score it set. Because the amount of shake was too high for the smaller scale images.

image

3. Replaced the action inside of OnCreated that was trying to set the text of the object to the value, with the custom action to change the score, because the previous action was not working, and didn't change the starting score set in the object properties.

image


HelperWesley:
In the latest commit I changed the origin point of the variants to 0,0 and made a change to the extension that's shown in the image below. I removed the default text action and added the custom rolling counter value action instead.

image

I don't really know why the default text action doesn't work, but using this action from the rolling counter behavior solves the issue and shows the value when the scene begins. So using this action is better than having an action inside of the extension that doesn't work. ⚖

PR #1014, opened 2026-06-25
New Babyfoot 3D free example by AnkhM - Créer Un jeu Vidéo

Bouh:


D8H:
It's a bit sad that this example ends with "3d" when every other 3D templates starts with "3d".

